PR #11, opened 2017-09-28
className can receive props values

cat2608:
This change makes possible use custom classNames via `props`, for instance, when working with scoped class names when the app doesn't relies on Bootstrap.
Our primary use case for this today is for our [Web Widget/Universal Loader](https://github.com/transferwise/widgets) where we want to ensure we have isolated styling because we cannot rely on client's Bootstrap version.


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.4.2
```
Tankenstein: This is more of an addition rather than a fix, so `v0.5.0`

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+## Accept className from `props`
+
+Use custom classNames via props:
+
+* Pass object `classes` with classes name: `classes={{ active: "active_3Ur-d" }}`
```
Tankenstein: I would call it `classNames` instead

Tankenstein:
Where did the yarn lockfile changes come from?

cat2608:
@Tankenstein After running `yarn` these dependencies were updated. Should I revert?

cat2608:
Updated docs:

screen shot 2017-09-29 at 11 48 38

@coolsebz @Tankenstein is it understandable?

PR #16, opened 2017-10-15
Add missing className as props to Option

cat2608:
ClassName as props to `<Option />` component. Changes missed on PR #11


package.json
```diff
@@ -39,12 +39,12 @@
     "gh-pages": "^0.12.0",
     "html-webpack-plugin": "2.24.0",
     "http-proxy-middleware": "0.17.3",
-    "jest": "18.1.0",
+    "jest": "21.2.1",
```
cat2608: I updated `jest` due to this error https://circleci.com/gh/transferwise/components/117 (same in local) and following this thread https://github.com/facebook/jest/issues/3721

package.json
```diff
@@ -57,6 +57,7 @@
   },
   "dependencies": {
     "currency-flags": "git+https://github.com/transferwise/currency-flags.git",
+    "jest-cli": "^21.2.1",
```
Tankenstein: Could we move all dev dependencies to devdependencies? Like jest cli and less and so on

PR #17, opened 2017-10-17
Smaller stepper labels

Tankenstein:
Makes stepper labels smaller.

Before:

image

After:

image


StevePole:
Think this should have been a Bootstrap change, not markup...

PR #18, opened 2017-10-23
Hover labels hover state

Tankenstein:
Not hovering:

image

Hovering:

image

Plus refactor a tiny bit


src/stepper/Stepper.less
```diff
@@ -12,13 +12,18 @@
...
       font-weight: @font-weight-semi-bold;
     }
 
     &--clickable .tw-stepper__step-label {
-      color: #1bc0ff;
+      color: @brand-light-blue;
       font-weight: @font-weight-semi-bold;
+      transition: color 0.15s ease-in-out;
+
+      &:hover {
```
oliverviljamaa: But what about `:focus`?

StevePole:
Why do you need these styles, why can't you use Bootstrap?

StevePole:
If for some reason you can't use Bootstrap, there are better variables to use so that you stay consistent:
image

PR #19, opened 2017-10-25
Add native react Tooltip component

Tankenstein:
And make the `Stepper` use it. 

image

image


src/stepper/Stepper.less
```diff
@@ -77,57 +35,14 @@
...
-    content: '';
-    display: block;
-    position: absolute;
 
-    margin-left: auto;
-    margin-right: auto;
-    left: 0;
-    right: 0;
-    top: -8px;
+  & .tooltip {
```
oliverviljamaa: Don't you want to namespace this? Otherwise some external styles could apply to this, `tooltip` is pretty general.

Tankenstein: It's a tooltip inside of a Stepper. The only place where a tooltip makes sense inside of a stepper is the label, where it's already defined by the component

oliverviljamaa: Oh, it's the Bootstrap class name, got it then.

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,112 @@
...
+        top: `${(parent.offsetTop + (parent.offsetHeight / 2)) - (tooltip.offsetHeight / 2)}px`,
+        left: `${(parent.offsetLeft + parent.offsetWidth) + offset}px`,
+      };
+    case Position.BOTTOM:
+      return {
+        top: `${(parent.offsetTop + tooltip.offsetHeight) + offset}px`,
+        left: `${(parent.offsetLeft + (parent.offsetWidth / 2)) - (tooltip.offsetWidth / 2)}px`,
+      };
+    default:
+      return {};
```
coolsebz: Why not add a sensible default? 🤔

PR #21, opened 2017-12-06
Stepper: Navigation between visited steps

wilsonjs:
back-and-forth


docs/StepperDocs.js
```diff
@@ -1,7 +1,39 @@
 import React, { Component } from 'react';
 import { Stepper, Select } from '../src';
 
-const presets = [
+const getBackAndForthSteps = (component, highestVisitedStep) => {
```
wilsonjs: Function that generates the step list, I'm passing in the component to get access to the state but ideally this whole thing would be inside the component itself having direct access to state.

wilsonjs: Function that generates the step list. on line 17 each step gets attached the onClick event handler if it has been visited or not.

docs/StepperDocs.js
```diff
@@ -1,7 +1,39 @@
...
+
+  return steps;
+};
+
+function goToBackAndForthStep(step, component) {
+  const index = step.index === undefined ? step.value : step.index;
+  const highestVisitedStep = Math.max(index, component.state.highestVisitedStep);
+
+  const preset = component.state.preset;
+  preset.steps = getBackAndForthSteps(component, highestVisitedStep);
```
wilsonjs: n order for the Stepper to know about the visited steps, we need to toggle the `onClick` prop, hence, the step list needs to be regenerated.

PR #24, opened 2017-12-19
Fix stepper tooltips width and position

KarlEST:
# Before:
image
image

# After:
image
image

Tooltips in stepper now have max-width: 100%.
On sm and md screens first and last tooltips are moved more to centre so they aren't offscreen.

https://transferwise.atlassian.net/browse/CON-966
After discussing with Helena we decided to go with not wrapping the words but instead increase the tooltip size.


src/stepper/Stepper.less
```diff
@@ -26,6 +26,20 @@
         color: @brand-light-blue-minus-10;
       }
     }
+
+    @media screen and (min-width: @screen-sm-min) and (max-width: @screen-md-max) {
+      &:first-child .tooltip-inner{
+        transform: translateX(~'calc(50% - 32px)');
```
mstaicu: What is the tilde doing? This is the first time I've seen it used in a translate.

wilsonjs: it's to escape the expression from `less`, otherwise less might try to do the calculation itself.

PR #25, opened 2018-01-17
Make stepper tooltips only appear on non-touch devices

Tankenstein:
Tooltips on the stepper on touch devices have caused problems with going back and overall confusion because of their behaviour, so we removed them.


docs/StepperDocs.js
```diff
@@ -134,8 +140,9 @@ export default class StepperDocs extends Component {
               }))}
               onChange={value => this.goToStep(value)}
               selected={{
-                label: `${this.state.activeStep} - ${this.state.selectedPreset.steps[this.state.activeStep]
-                  .label}`,
+                label: `${this.state.activeStep} - ${
```
OFFLlNE: interesting prettier :)

oliverviljamaa: A bit weird, but better than before 😀

src/stepper/deviceDetection.js
```diff
@@ -3,16 +3,18 @@ function supportsTouchEvents() {
...
+    'iemobile',
+    'iphone',
+    'ipad',
+    'ipod',
+    'blackberry',
+    'bada',
+  ];
+  const matchString = sampleTouchDevices.map(device => `(${device})`).join('|');
+  const regex = new RegExp(matchString, 'ig');
+  return !!navigator.userAgent.match(regex);
```
mstaicu: I would go with `Boolean(..)` for readability.

oliverviljamaa: I think `!!` is more popular and therefore better recognized?

PR #28, opened 2018-02-13
Select fixes and tests

czebe:
* Open Select by pressing UP/DOWN arrows
* Open Select by pressing SPACE
* Select an option by pressing SPACE
* Disable TAB navigation when Select is open, mimicking default behaviour
* Close Select when clicking on another one (no more multiple Selects open)
* Attach document event listeners only when Select is open
* Add tests for the functionalities above
* Extract helper utilities to `common` directory to be shared by specs
* Simplify existing tests for better readability


package.json
```diff
@@ -98,24 +102,35 @@
...
   },
   "babel": {
-    "presets": ["react-app"]
+    "presets": [
+      "react-app"
+    ]
   },
   "lint-staged": {
-    "*.js": ["prettier-eslint '**/*.js' --write", "git add"]
+    "*.js": [
```
oliverviljamaa: Why the formatting change? Do you have Prettier on in your IDE?

src/select/Select.spec.js
```diff
@@ -107,16 +65,28 @@ describe('Select', () => {
     expectDropdownToBe().open();
   });
 
+  it('can be opened by UP/DOWN arrows and SPACE', () => {
```
oliverviljamaa: Could we have separate `it` blocks for these?

src/select/Select.spec.js
```diff
@@ -285,50 +255,68 @@ describe('Select', () => {
...
-    doTimes(103, () => component.simulate('keyDown', fakeKeyDownEventForKey(KeyCodes.UP)));
-    expect(findNthListElement(0).hasClass('active')).toBe(true);
-    expect(findNthListElement(1).hasClass('active')).toBe(false);
-    doTimes(103, () => component.simulate('keyDown', fakeKeyDownEventForKey(KeyCodes.DOWN)));
-    expect(findNthListElement(0).hasClass('active')).toBe(false);
-    expect(findNthListElement(1).hasClass('active')).toBe(true);
+
+    // Move to the bottom of Select with 3 options
+    doTimes(3, () => component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.DOWN)));
+    expect(element('.active').length).toBe(1);
```
oliverviljamaa: Could we do `expect(activeOptionIndex()).toBe(2)` here instead to not repeat ourselves that much and encapsulate the knowledge about the `.active`?

src/select/Select.spec.js
```diff
@@ -285,50 +263,64 @@ describe('Select', () => {
...
-  it('allows you to select the item currently focused with your keyboard', () => {
-    const onChange = jest.fn();
-    component.setProps({
-      options: [{ value: 0, label: 'yo' }, { value: 1, label: 'dawg' }, { value: 2, label: 'boi' }],
-      required: true,
-      onChange,
-    });
-    doTimes(2, () => component.simulate('keyDown', fakeKeyDownEventForKey(KeyCodes.DOWN)));
-    // hitting it once will only make the selector appear, so we hit it twice to go down once.
+  it('allows you to select the item currently focused with your keyboard by pressing ENTER or SPACE', () => {
```
oliverviljamaa: I'd split the ENTER and SPACE checks to two separate tests as well, WDYT?

czebe: True!

PR #30, opened 2018-02-15
Bump node version

Tankenstein:
Components release job failed, probably due to old node not supporting trailing comma.


circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 6.9.1
+    version: 8.9.1
```
lapostoj: 8.9.4 ? 🙂

Tankenstein: sure?

Tankenstein:
Will squash commits

PR #32, opened 2018-03-02
Improve flow navigation

lapostoj:


lapostoj:
Yep, to be able to use on the "done" screens and have something like that
image

Tankenstein:
@lapostoj @oliverviljamaa I'm wondering if it would be nicer to define a constant (oo maybe a symbol) that we export, like 

```jsx
<FlowNavigation activeStep={FlowNavigation.DONE} />
```

lapostoj:
I was thinking about it but could it be that it won't always be in a "DONE" context? 
In itself the component doesn't have to know anything about that I would say 🤔 

Also extra small one, objections to hide the bottom border the same way as the one near the "x"?

Tankenstein:
Can you also add something about this to docs?

src/flowNavigation/FlowNavigation.js
```diff
@@ -23,8 +23,8 @@ const BackArrow = () => (
   </svg>
 );
 
-const FlowNavigation = ({ steps, activeStep, avatarUrl, profileType, onClose, onGoBack }) => (
-  <div className="tw-flow-navigation">
+const FlowNavigation = ({ steps, activeStep, avatarUrl, done, profileType, onClose, onGoBack }) => (
+  <div className={`tw-flow-navigation ${done ? 'tw-flow-navigation__done' : ''}`}>
```
Tankenstein: Since it's a modifier, use `--done` not `__done`

PR #33, opened 2018-03-08
Select UX improvements

czebe:
Select UX improvements

- Jump to the first matching option when user starts typing in an open `Select`
- Allow navigation by arrows even if user is typing into the search field
- Clear search field when `Select` is closed
- Upgrade Jest to latest version


src/select/Select.js
```diff
@@ -130,6 +142,32 @@ export default class Select extends Component {
         }
         break;
       default:
+        if (event.keyCode === KeyCodes.SPACE && this.state.keyboardInputSelection === '') {
```
andrastothtw: I would encapsulate this in a function as this is a huge block of code.

src/select/Select.js
```diff
@@ -130,6 +142,32 @@ export default class Select extends Component {
...
+          if (open) {
+            this.selectKeyboardFocusedOption();
+          } else {
+            this.open();
+          }
+        } else if (/\w|\s/i.test(key)) {
```
andrastothtw: Please name your regexes. For example: `const anyNonSymbols = /\w|\s/i;`

src/select/Select.js
```diff
@@ -130,6 +142,32 @@ export default class Select extends Component {
...
+          this.setState((previousState, previousProps) => {
```
andrastothtw: I would also name this function, since it's longer than a couple of lines and does something specific.

src/select/Select.js
```diff
@@ -130,6 +142,32 @@ export default class Select extends Component {
...
+              unClampedNewIndex > -1
+                ? clamp(0, optionsWithoutHeaders.length - 1, unClampedNewIndex)
+                : previousState.keyboardFocusedOptionIndex;
+
+            return { keyboardInputSelection: keyboardInput, keyboardFocusedOptionIndex: newIndex };
+          });
+          clearTimeout(this.keyboardInputSelectionTimeout);
+          this.keyboardInputSelectionTimeout = setTimeout(() => {
+            this.setState({ keyboardInputSelection: '' });
+          }, 1000);
```
andrastothtw: What is this timeout doing?

src/select/Select.js
```diff
@@ -181,8 +223,9 @@ export default class Select extends Component {
     }
   };
 
-  handleDocumentClick = () => {
-    if (this.state.open) {
+  handleDocumentClick = event => {
+    // Ignore click events triggered by hitting SPACE
```
andrastothtw: Please add more details into the comment on why testing with `screenX` and `screenY` filter `SPACE` events.

src/select/Select.spec.js
```diff
@@ -346,4 +344,63 @@ describe('Select', () => {
...
+    component.setProps({ required: true });
+    openSelect();
+
+    expect(activeOptionIndex()).toBe(null);
+
+    component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.d));
+    expect(activeOptionIndex()).toBe(1);
+  });
+
+  it('highlights the matching option on keyboard input with multiple characters', () => {
```
andrastothtw: Nice, clear testing code! 👍

src/select/Select.spec.js
```diff
@@ -346,4 +344,63 @@ describe('Select', () => {
...
+    });
+    openSelect();
+
+    component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.a));
+    expect(activeOptionIndex()).toBe(0);
+
+    component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.b));
+    expect(activeOptionIndex()).toBe(0);
+
+    jest.advanceTimersByTime(1100);
```
andrastothtw: For search timeouts I have found 300-500 ms working smoother. Especially if you don't need to care about API calls.

src/common/keyCodes.js
```diff
@@ -5,4 +5,30 @@ export default {
   ESCAPE: 27,
   TAB: 9,
   SPACE: 32,
+  a: 65,
```
Tankenstein: Why add this?

src/select/Select.js
Tankenstein: ☝️

src/select/Select.js
Tankenstein: ☝️

src/select/Select.js
```diff
@@ -89,8 +95,22 @@ export default class Select extends Component {
 
   handleSearchChange = event => this.props.onSearchChange(event.target.value);
 
+  handleSearchKeydown = event => {
+    if (
+      event.keyCode !== KeyCodes.UP &&
```
Tankenstein: Would be nicer to use an `indexOf` here i think

src/select/Select.js
```diff
@@ -89,8 +95,22 @@ export default class Select extends Component {
...
+      event.keyCode !== KeyCodes.ENTER &&
+      event.keyCode !== KeyCodes.TAB
+    ) {
+      event.stopPropagation();
+    }
+  };
+
   handleKeyDown = event => {
     const { open } = this.state;
+    const key = event.key;
```
Tankenstein: Could you explain why you destructure this here?

src/select/Select.js
Tankenstein: I don't get what keyboardInputSelection is used for either

src/select/Select.spec.js
```diff
@@ -221,28 +219,28 @@ describe('Select', () => {
...
+    openSelect();
     expect(component.find('input').prop('placeholder')).toEqual('hello');
   });
 
   it('does not show placeholder option when search enabled', () => {
     component.setProps({ placeholder: 'ayy lmao' });
     openSelect();
     expect(component.find('ul').text()).toContain('ayy lmao');
-    openSearchableSelect();
+    component.setProps({ onSearchChange: jest.fn() });
```
Tankenstein: Why remove these abstractions? Reads quite nicely imo

src/select/Select.spec.js
Tankenstein: @andrastothtw this is sync, there should be no timeouts :/

czebe: Timeouts -- imagine the following select:

```
<select>
  <option>HI</option>
  <option>HELLO</option>
  <option>ECHO</option>
  <option>LOLO</option>
</select>
```

When you open a select and start typing, it jumps to the first match. You can add multiple characters within a certain timeframe (hence the timeout). Suppose:

You type:  H jumps to HI
You type:  H-E-L jumps to HELLO
If you go H.....wait.....E jumps to ECHO since the timeout has expired between the two characters.

Timeout for waiting for additional characters is *1000ms*, as this feels natural and close to the native browser implementation.

czebe:
Wait, this is not a search functionality, this is related to how Select works natively in the browser. You can press keyboard keys and it jumps to a matching option. I don't think this should be an 'add-on', it should be the default behaviour of every Select, as users are expecting it.

dozortsev:
Not relevant any more.

PR #34, opened 2018-03-15
Stop bundling flags

oliverviljamaa:
# v2.0.0
## Stops bundling currency flags

This makes the output CSS smaller and means we can avoid duplicating the flags CSS
if our app needs it in other places as well.


oliverviljamaa:
Good point, @Tankenstein. Also, are you fine with the fact that it’s a major bump?

oliverviljamaa:
@Tankenstein Added to `devDependencies` and made sure it works in the docs.

Tankenstein:
Yep this is backwards incompatible.

oliverviljamaa:
@Tankenstein The file size difference is as follows:

**Before:**: 39.87 kB
**After:** 1.23 kB

Tankenstein:
@oliverviljamaa wait, seriously?

Tankenstein:
fuckin' a man

PR #40, opened 2018-05-04
Adds a new money input component.

Tankenstein:
image

Also adds a "separator" element and a bunch of dropdown utilities to the select component.

### DIFF
--- a/src/moneyInput/MoneyInput.js
+++ b/src/moneyInput/MoneyInput.js
@@ -1,0 +1,104 @@
+import React, { Component } from 'react';
+import Types from 'prop-types';
+import Select from '../select';
+
+const Currency = Types.shape({
+  header: Types.string,
+  value: Types.string,
+  label: Types.string,
+  currency: Types.string,
+  note: Types.string,
+  searchable: Types.string,
+});
+
+class MoneyInput extends Component {
+  static propTypes = {
+    currencies: Types.arrayOf(Currency).isRequired,
+    selectedCurrency: Currency.isRequired,
+    onCurrencyChange: Types.func.isRequired,
+    amount: Types.number.isRequired,
+    size: Types.oneOf(['xs', 'sm', 'md', 'lg']),
+    onAmountChange: Types.func.isRequired,
+  };
+
+  static defaultProps = {
+    size: 'lg',
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      searchQuery: '',
+    };
+  }
+
+  getSelectOptions() {
+    const headerMapping = this.props.currencies.reduce(
+      (accumulator, option) => {
+        if (option.header) {
+          accumulator.currentHeader = option.header; // eslint-disable-line no-param-reassign
+        } else if (option.value) {
+          accumulator.mapping[option.value] = // eslint-disable-line no-param-reassign
+            accumulator.currentHeader;
+        }
+        return accumulator;
+      },
+      { currentHeader: null, mapping: {} },
+    ).mapping;
+
+    const foundOptions = this.props.currencies.filter(option =>
+      this.includeCurrencyInSearchResults(option),
+    );
+
+    return this.props.currencies.filter(option => {
+      if (option.header) {
+        return foundOptions.reduce(
+          (headerVisible, foundOption) =>
+            headerVisible || headerMapping[foundOption.value] === option.header,
+          false,
+        );
+      }
+      return foundOptions.indexOf(option) !== -1;
+    });
+  }
+
+  includeCurrencyInSearchResults(currency) {
+    const searchQuery = this.state.searchQuery.toLowerCase();
+    return (
+      currency.label &&
+      (currency.label.toLowerCase().indexOf(searchQuery) !== -1 ||
+        (currency.searchable && currency.searchable.toLowerCase().indexOf(searchQuery) !== -1) ||
+        (currency.note && currency.note.toLowerCase().indexOf(searchQuery) !== -1))
+    );
+  }
+
+  render() {
+    const { selectedCurrency, amount, onAmountChange, onCurrencyChange, size } = this.props;
+    // TODO: amount handling
+    return (
+      <div className={`input-group input-group-${size}`}>
+        <input
+          value={amount}
+          className="form-control"
+          onChange={event => onAmountChange(parseInt(event.target.value, 10))}
+        />
+        <span className="input-group-btn amount-currency-select-btn">
+          <Select
+            options={this.getSelectOptions()}
+            selected={{ ...selectedCurrency, note: null }}
+            onChange={onCurrencyChange}
+            onSearchChange={searchQuery => this.setState({ searchQuery })}
+            searchValue={this.state.searchQuery}
+            size={size}
+            required
+            dropdownRight="xs"
+            dropdownWidth="lg"
+            inverse
+          />
+        </span>
+      </div>
+    );
+  }
+}
+
+export default MoneyInput;
--- a/src/moneyInput/MoneyInput.js
+++ b/src/moneyInput/MoneyInput.js
@@ -188,7 +188,7 @@ class MoneyInput extends Component {
               ? [
                 <i className="tw-money-input__keyline" key="keyline" />,
                 <i
-                  className={`currency-flag currency-flag-${selectedCurrency.currency.toLowerCase()} hidden-xs`}
+                  className={`currency-flag currency-flag-${selectedCurrency.currency.toLowerCase()} hidden-xs m-r-2`}
                   key="flag"
                 />,
                 ]
--- a/src/moneyInput/MoneyInput.js
+++ b/src/moneyInput/MoneyInput.js
@@ -17,7 +17,7 @@ class MoneyInput extends Component {
     selectedCurrency: Currency.isRequired,
     onCurrencyChange: Types.func.isRequired,
     amount: Types.number.isRequired,
-    size: Types.oneOf(['xs', 'sm', 'md', 'lg']),
+    size: Types.oneOf(['sm', 'md', 'lg']),
     onAmountChange: Types.func.isRequired,
   };
 
--- a/src/moneyInput/MoneyInput.js
+++ b/src/moneyInput/MoneyInput.js
@@ -19,36 +19,32 @@ class MoneyInput extends Component {
     id: Types.string,
     currencies: Types.arrayOf(Currency).isRequired,
     selectedCurrency: Currency.isRequired,
-    onCurrencyChange: Types.func.isRequired,
+    onCurrencyChange: Types.func,
     amount: Types.number.isRequired,
     size: Types.oneOf(['sm', 'md', 'lg']),
-    onAmountChange: Types.func.isRequired,
-    numberFormatLocale: Types.string,
+    onAmountChange: Types.func,
+    locale: Types.string,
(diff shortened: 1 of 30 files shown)


src/moneyInput/numberFormatting.js
```diff
@@ -0,0 +1,32 @@
+function isNumberLocaleSupported() {
```
lapostoj: can be done later but maybe that should be in it's own folder rather than nested in moneyInput, could be useful in itself as a formatter.

src/select/Select.js
```diff
@@ -27,7 +27,10 @@ export default class Select extends Component {
     id: Types.string,
     required: Types.bool,
     disabled: Types.bool,
-    size: Types.oneOf(['xs', 'sm', 'md', 'lg']),
+    inverse: Types.bool,
+    dropdownRight: Types.oneOf(['xs', 'sm', 'md', 'lg', 'xl']),
```
lapostoj: I know it's in the angular version as well but what does it do again? 😅

src/select/Select.less
```diff
@@ -8,4 +8,8 @@
     border-bottom-width: 1px;
     border-bottom-color: #f2f5f7;
   }
+
+  .currency-flag {
+    margin-right: 8px;
```
lapostoj: shouldn't that just be a `m-r-1`?

src/moneyInput/MoneyInput.less
```diff
@@ -0,0 +1,33 @@
...
+  transform: translateY(4px);
+  }
+
+  &--disabled {
+    background-color: @brand-smoke-plus-20;
+    cursor: not-allowed;
+  }
+
+  &__fixed-currency {
+    .currency-flag {
```
lapostoj: is 14px ok? if 16, shouldn't that just be a `m-r-2`?

src/select/Select.js
Tankenstein: You set a breakpoint, from that breakpoint it'll start opening from the other side. Only visible if not a block button, like in case of the input addon select component.

src/moneyInput/numberFormatting.js
Tankenstein: Would move it once someone needs it, not before.

src/moneyInput/MoneyInput.js
```diff
@@ -0,0 +1,222 @@
...
+  static propTypes = {
+    id: Types.string,
+    currencies: Types.arrayOf(Currency).isRequired,
+    selectedCurrency: Currency.isRequired,
+    onCurrencyChange: Types.func.isRequired,
+    amount: Types.number.isRequired,
+    size: Types.oneOf(['sm', 'md', 'lg']),
+    onAmountChange: Types.func.isRequired,
+    numberFormatLocale: Types.string,
+    numberFormatPrecision: Types.number,
```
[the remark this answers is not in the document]
Tankenstein: This component shouldn't know anything about the details of a currency imo.

src/moneyInput/MoneyInput.less
```diff
@@ -0,0 +1,33 @@
+@import '~bootstrap/less/variables';
+
+.tw-money-input {
+  .input-group-addon {
+    padding-left: 0 !important;
+  }
+
+  &__keyline {
+    width: 2px;
```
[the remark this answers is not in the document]
Tankenstein: Yep ty

src/moneyInput/MoneyInput.less
Tankenstein: In the mocks i believe it as actually 14, will double check and maybe still change to m-r-2

src/moneyInput/MoneyInput.js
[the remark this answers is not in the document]
lapostoj: But shouldn't that be expected in the "currency" value?
Don't remember if in angular it works like this or if the formatting know about the decimals, (from memory I would say 2nd option) but having to deal with it as a user of the component sounds painful 😅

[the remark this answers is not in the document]
Tankenstein: Well, we should be using the same formatting we get from the api, not automatic generic stuff imo. This supports that. However, defo something to think about, but right now it works for the place where we'll use it.

src/select/Select.less
Tankenstein: Should be, but we're style sniping here because some styles differ from the "breakdown" variant to the "selected" variant, while they both use the same component to render

src/moneyInput/MoneyInput.less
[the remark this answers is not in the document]
Tankenstein: (fixed now)

src/moneyInput/MoneyInput.js
```diff
@@ -0,0 +1,222 @@
...
+    addon: Types.node,
+    disabled: Types.bool,
+    searchPlaceholder: Types.string,
+  };
+
+  static defaultProps = {
+    id: undefined,
```
skywritergr: Defining things as undefined yourself is never a good idea. It's always better to keep it to `null` instead. Is there a specific reason for this that I'm missing?

Tankenstein: There is, and i agree with you. I didn't want to pass a "null" id to an actual dom element, where this will end up.

src/moneyInput/numberFormatting.js
```diff
@@ -0,0 +1,32 @@
...
+  const number = 1234;
```
skywritergr: Again, it might be that I'm totally ignorant, but why are we doing this? I wouldn't mind a small comment around the area describing this functionality.

src/select/Select.js
```diff
@@ -81,7 +88,7 @@ export default class Select extends Component {
 
   getIndexWithoutHeadersForIndexWithHeaders(index) {
```
skywritergr: Can we have a bit less descriptive function names? 😅

Tankenstein: 🙅

src/moneyInput/numberFormatting.js
Tankenstein: This is functionality to parse and render numbers in a given locale and precision. For example, in estonia, i would write 1234.56 like so: 1 234,56 while in the uk you would go 1,234.56

src/moneyInput/MoneyInput.js
```diff
@@ -0,0 +1,222 @@
...
+  note: Types.string,
+  searchable: Types.string,
+});
+
+class MoneyInput extends Component {
+  static propTypes = {
+    id: Types.string,
+    currencies: Types.arrayOf(Currency).isRequired,
+    selectedCurrency: Currency.isRequired,
+    onCurrencyChange: Types.func.isRequired,
```
coolsebz: hmm, why is this always required? (from an API design POV)

coolsebz: same question for `onAmountChange` too

src/moneyInput/MoneyInput.js
oliverviljamaa: Are you sure? Some attributes are omitted when their values are falsy.

src/moneyInput/MoneyInput.js
Tankenstein: Actually solid point. From an API standpoint, what makes sense to us is that

When you have falsy onCurrencyChange, this also means you should see the "fixed currency" input
and
When you have falsy onAmountChange, this also means the input should be disabled.

However, now we have essentially no need for the "disabled" prop. Should we remove it?

StevePole: Had the same question but hadn't posted review yet.  My perspective was that it should be possible to use forms with native submit behaviour, in which case no handler is necessary.  Regards `disabled` I would think about consistency across all components, if that is the pattern they all follow it's fine.  If that's something specific to this control it's not so good.  To me feels clearer to have it separate and not assume that the lack of handler implies it should be disabled.

src/moneyInput/MoneyInput.js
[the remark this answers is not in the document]
StevePole: I'm more in Jerome's camp, it makes it harder to use the component (correctly), and pretty straightforward data to keep in the client, slow rate of change in the underlying data. https://github.com/transferwise/styleguide-components/blob/master/src/services/currency/currency.service.js

src/moneyInput/MoneyInput.js
coolsebz: yeah would make sense to remove it then if the intended use is with a falsy onAmountChange 👌

src/moneyInput/MoneyInput.spec.js
```diff
@@ -0,0 +1,373 @@
...
+    expect(currencySelect().prop('selected')).toEqual({
+      value: 'EUR',
+      label: 'EUR',
+      note: null,
+      currency: 'eur',
+      searchable: 'Spain, Germany, France, Austria, Estonia',
+    });
+  });
+
+  it('calls onCurrencyChange when the user selects a different currency', () => {
```
StevePole: This is a stylistic point, feel free to ignore.  I've taken to formulate my tests using a 'Given, when, then' format which (for me) helps keep them behavioural, groups the same behaviours and makes it (imo) more readable.  So I never put a 'when' in the assertion clause, only the resulting behaviour.  The test setup that creates the 'when' scenario can then naturally have it's own beforeEach, and the assertion clauses just contain assertions.  In this case, you would effectively have:
```
describe('given a MoneyInput', () => {
  ...
  describe('when the user selects a different currency', () => {
    beforeEach(() => {
      // simulate selecting different currency
    });
    it('should call onCurrencyChange', () => {
      ...
    });
    it('should update the selected currency', ...);
  });
});
```

src/moneyInput/numberFormatting.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+    expect(formatNumber(1234.56, 'en-GB', 3)).toBe('1,234.560'); // sanity check
+  });
+
+  it('uses toFixed to format if localeString not supported or acts weirdly', () => {
+    const toLocaleString = Number.prototype.toLocaleString;
+    // eslint-disable-next-line no-extend-native
+    Number.prototype.toLocaleString = null;
+
+    expect(formatNumber(fakeNumber, 'en-GB', 5)).toBe('fixed for precision 5');
```
StevePole: Found this approach to the test a bit confusing.  Do you actually need the lines above?  I don't think your testing a real number, so that should be irrelevant?

src/select/Select.js
```diff
@@ -10,8 +10,8 @@ function clamp(from, to, value) {
   return Math.max(Math.min(to, value), from);
 }
 
-function notHeader(option) {
-  return !option.header;
+function actionableOption(option) {
```
StevePole: `isActionable`?

src/moneyInput/MoneyInput.js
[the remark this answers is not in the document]
Tankenstein: After seeing how simple that service is, i tend to agree. We'll change.

src/moneyInput/MoneyInput.spec.js
Tankenstein: Fair point, i follow this as well sometimes, but tend to think a bit too little about test readability usually.

src/moneyInput/numberFormatting.spec.js
Tankenstein: You are correct, was probably changed at some point without observing the context.

PR #44, opened 2018-05-15
Adds inverse state to FlowNavigation

williambaumgartner:
* Adds dark mode for FlowNavigation
* Adds new property `inverse` which shows the dark version
* Properly hides the border when `done` is `true` in FlowNavigation
* Adds `classnames` dev-dependency for easy class management
* Minor less refactoring

image


StevePole:
Ok, makes sense, but this is an open source project, it doesn't know anything about the batch flow and whether there will/won't be steps.  I would say if you're going to add the feature, you should make it work for all cases.

Would be great to see the designs for this flow if they're kicking around somewhere?

czebe:
You don't even need a FlowNavigation here, just a logo and a close button.

williambaumgartner:
@StevePole Makes sense. Will ask for designs and update accordingly.

williambaumgartner:
Closing this as I'll replicate the inverse done state.

PR #49, opened 2018-05-21
Add previous step name to back button for flow navigation

oliverviljamaa:
Issue: https://transferwise.atlassian.net/browse/CON-1068

# v3.1.0
## Add previous step name to back button for flow navigation

flow_navigation


src/flowNavigation/FlowNavigation.spec.js
```diff
@@ -26,11 +27,7 @@ describe('Flow navigation', () => {
   }
 
   function backButton() {
-    return component.find('.tw-flow-navigation__back-button');
-  }
-
-  function backButtonHidden() {
-    return backButton().hasClass('tw-flow-navigation__back-button--hidden');
+    return component.find(BackButton);
```
codewhisperer: Would it be cleaner if you added `!!` before?

Tankenstein: No - it's used to actually find the component and do stuff with it, not do a truthy check

src/flowNavigation/backButton/BackButton.js
```diff
@@ -0,0 +1,86 @@
...
+
+  render() {
+    const { steps, activeStep, onGoBack } = this.props;
+
+    return (
+      <button
+        className={`btn-unstyled visible-xs tw-flow-navigation__back-button ${
+          !onGoBack ? 'tw-flow-navigation__back-button--hidden' : ''
+        }`}
+        onClick={() => onGoBack && onGoBack()}
```
codewhisperer: I don't get this one. Is the first condition checking if the function is there?

src/flowNavigation/FlowNavigation.spec.js
codewhisperer: But shouldn't the function have a different name then?

codewhisperer: Oh never mind, thought the function name was backButtonHidden.

PR #55, opened 2018-07-11
HOM-29: Add custom action to money input

KarlEST:
image


src/moneyInput/MoneyInput.spec.js
```diff
@@ -338,4 +363,35 @@ describe('Money Input', () => {
...
+      ...currencies,
+      { value: 'CUSTOM_ACTION', label: 'Label boy' },
+    ]);
+    expect(displayedCurrencies()[displayedCurrencies().length - 1]).toEqual({
+      value: 'CUSTOM_ACTION',
+      label: 'Label boy',
+    });
+  });
+
+  it('calls onCustomAction callback when custom action is selected', () => {
```
OFFLlNE: I'd add some blank lines before expect-s for some better readability

docs/MoneyInputDocs.js
```diff
@@ -136,6 +137,10 @@ export default class MoneyInputDocs extends Component {
                   }
                   addon={this.state.addonEnabled ? <i className="icon icon-unlock" /> : undefined}
                   searchPlaceholder={this.state.searchPlaceholder}
+                  onCustomAction={
```
Tankenstein: Let's call it `onOtherCurrencySelected` maybe?

docs/MoneyInputDocs.js
```diff
@@ -136,6 +137,10 @@ export default class MoneyInputDocs extends Component {
...
+                    this.state.customActionEnabled ? () => alert('Custom action') : null // eslint-disable-line no-alert
+                  }
+                  customActionLabel={this.state.customActionEnabled ? 'Custom action label' : null}
```
Tankenstein: `otherCurrencyLabel`

docs/MoneyInputDocs.js
```diff
@@ -254,6 +261,15 @@ export default class MoneyInputDocs extends Component {
                 }
               />
 
+              <div className="m-t-3" />
+              <Checkbox
+                label="Custom action"
+                checked={this.state.customActionEnabled}
+                onChange={() =>
+                  this.setState({ customActionEnabled: !this.state.customActionEnabled })
```
Tankenstein: Can't do this. To set state derived from previous state, you have to use the function form, otherwise you'll hit race conditions because it's async.

`this.setState(oldState => ({ customActionEnabled: !oldState.customActionEnabled }))`

docs/MoneyInputDocs.js
KarlEST: I'm fine with this. WDYT @oliverviljamaa ?

oliverviljamaa: I prefer `onCustomAction`, as I wouldn't want to limit this too much – in `calculator`, I'd prefer `onOtherCurrencySelected`, though.

src/moneyInput/MoneyInput.js
```diff
@@ -80,7 +85,16 @@ export default class MoneyInput extends Component {
   };
 
   getSelectOptions() {
-    return filterOptionsForQuery(this.props.currencies, this.state.searchQuery);
+    const selectOptions = filterOptionsForQuery(
```
oliverviljamaa:
```js
const selectOptions = [...filterOptionsForQuery(
  this.props.currencies,
  this.state.searchQuery,
)];
```

PR #56, opened 2018-07-16
Fix npm tokens in circle config

KarlEST:


KarlEST:
@matallo Should be good now

bobsantos:
bitmoji

PR #58, opened 2018-07-31
Add popover component

oliverviljamaa:
# v4.1.0
## Add popover component

popover


src/popover/Popover.js
```diff
@@ -0,0 +1,80 @@
...
+    this.setState({ isOpen: false });
+    document.removeEventListener('click', this.closePopoverOnOutsideClick, true);
+  };
+
+  render() {
+    const { children, title, content, preferredPlacement } = this.props;
+    const { isOpen } = this.state;
+
+    const placement = getPlacement(this.popoverElement, preferredPlacement);
+    const popoverClassName = classNames(
```
codewhisperer: 😍

StevePole:
Is this the current animation?  Feels both a bit slow and a bit 'long'.  The animation for the select boxes is better, it provides the impression of direction without animating the entire size.  They definitely feel like they're coming from a different place.

oliverviljamaa:
@StevePole It is, and I completely agree with you, it's too intense and slow.

docs/PopoverDocs.js
```diff
@@ -0,0 +1,180 @@
...
+  <button className="btn btn-default">
+    Click me
+  </button>
+</Popover>`}
+        </pre>
+        {/* eslint-enable react/jsx-indent */}
+      </Fragment>
+    );
+
+    const contentControl = (
```
codewhisperer: Would it make sense to put those fragments into separate files?

docs/PopoverDocs.js
```diff
@@ -0,0 +1,180 @@
...
+    const { title, content, preferredPlacementKey, rightTrigger } = this.state;
+
+    const preferredPlacementKeys = Object.keys(Popover.Placement);
+    const Code = () => (
+      <Fragment>
+        {/* eslint-disable react/jsx-indent */}
+        <pre className="tw-docs-code">
+          {`<Popover${
+            title
+              ? `
```
codewhisperer: I'm guessing you already tried to find a nicer way of doing this. It would be great if react would just remove the attribute if it was false or null

docs/PopoverDocs.js
```diff
@@ -0,0 +1,180 @@
...
+      </Fragment>
+    );
+
+    const titleControls = (
+      <div className="row m-t-2">
+        <div className="col-md-6">
+          <Checkbox
+            label="Title"
+            onChange={() => {
+              if (title) {
```
codewhisperer: can you use a ternary here?

docs/PopoverDocs.js
oliverviljamaa: I don't know – I thought about it and all the other docs are just one long file as well, I'd be the first one to introduce a folder.

docs/PopoverDocs.js
oliverviljamaa: I'm not really returning anything here and ESLint doesn't allow floating expressions.

oliverviljamaa: Well, React does, but this is just string manipulation.

docs/PopoverDocs.js
oliverviljamaa: I'm not returning anything here and ESLint doesn't allow floating expressions.

docs/PopoverDocs.js
Tankenstein: I would keep the docs simple and nasty like this.

src/popover/positioning.js
```diff
@@ -0,0 +1,101 @@
...
+        left: triggerLeft - popoverWidth,
+      };
+    default:
+      return {};
+  }
+}
+
+function toInt(str) {
+  return parseInt(str, 10);
+}
```
Tankenstein: I think you did a really lovely job here, damn

src/stepper/deviceDetection.js
```diff
@@ -22,7 +22,7 @@ function userAgentSuggestsTouchDevice() {
 }
 // Important: this is not fool-proof! It gives false positives and negatives, and will be outdated.
 // Only use this for small vanity changes where it still works either way.
-// eslint-disable-next-line import/prefer-default-export
+// eslint-disable-next-line
```
Tankenstein: Why disable it if the rule is disabled now?

oliverviljamaa: No reason, just a dumb find-replace!

src/popover/positioning.js
```diff
@@ -0,0 +1,101 @@
...
+  switch (placement) {
+    case TOP:
+      return {
+        top: triggerTop - popoverHeight,
+        left: triggerLeft + triggerWidth / 2 - popoverWidth / 2,
+      };
+    case RIGHT:
+      return {
+        top: triggerTop + triggerHeight / 2 - popoverHeight / 2,
+        left: triggerLeft + triggerWidth,
```
mstaicu: Would be nice to have a `margin` property, as a future improvement, passed to the component that will add a given space amount for RIGHT and LEFT placements.

src/popover/Popover.js
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!popoverClicked) {
+      this.close();
+    }
+  };
+
+  open = () => {
+    this.setState({ isOpen: true });
+    document.addEventListener('click', this.closePopoverOnOutsideClick, true);
+  };
```
Tankenstein: Potential listener leak -> if the component is unmounted when it's open, it might keep the listener, worth a check

PR #63, opened 2018-08-06
Change internal popover markup

oliverviljamaa:
# v4.1.2
## Change internal popover markup

Previously, popover created a container element for the trigger passed as a child and the popover itself, and overrode styles to make the popover itself not be styled like the container that acted as a trigger.

Now, popover decorates the trigger passed as a child with accessibility props and an event handler, removing the need for a wrapper element and custom styles.

This also enables focus style support out of the box.


src/popover/Popover.js
```diff
@@ -47,27 +46,30 @@ export default class Popover extends Component {
...
   };
 
+  createTrigger = () => {
+    const { children: child } = this.props;
+
+    return cloneElement(typeof child === 'string' ? <span>{child}</span> : child, {
+      'data-toggle': 'popover',
+      role: 'button',
+      tabIndex: 0,
+      onClick: this.open,
```
KarlEST: no more `onKeyPress`?

src/popover/Popover.js
```diff
@@ -47,27 +46,30 @@ export default class Popover extends Component {
     document.removeEventListener('click', this.closePopoverOnOutsideClick, true);
   };
 
+  createTrigger = () => {
+    const { children: child } = this.props;
+
+    return cloneElement(typeof child === 'string' ? <span>{child}</span> : child, {
```
KarlEST: Not sure I understand correctly but doesn't `string` one need the `onClick` etc as well to open the popover?

oliverviljamaa: The second argument, the options object, is not part of the ternary, and is added whatever the first argument is.

src/popover/Popover.js
oliverviljamaa: No, `tabIndex` ensures it's keyboard navigatable.

src/popover/Popover.js
Tankenstein: Curious: Why do you need to put it in a span if it's a string?

oliverviljamaa: Sometimes the child is not a React element, but a string. In that case, I need a wrapper element for attaching keypress and click listeners.

Tankenstein: right, cloneelement does not work for strings 🤦‍♂️

PR #64, opened 2018-08-22
HOM-36: Add node support for custom label and Select labels

KarlEST:


src/moneyInput/MoneyInput.js
```diff
@@ -27,7 +27,7 @@ export default class MoneyInput extends Component {
     locale: Types.string,
     addon: Types.node,
     searchPlaceholder: Types.string,
-    customActionLabel: Types.string,
+    customActionLabel: Types.oneOfType([Types.string, Types.element]),
```
Tankenstein: Use `Types.node`

PR #65, opened 2018-08-22
HistoricRatesChart [WIP]

matallo:
- Added support for HistoricRatesChart

Extracted component from Homepage, to be used in Currency Converter

localhost_3000_


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v4.3.0
```
oliverviljamaa: Love that we're extracting this!

However, I'm not entirely sure whether it fits with the other components in this repo. These are mainly component implementations for our design system molecules, whereas this seems like something more specific (think `public-navigation` or `calculator`, f.e.).

What would you think about having it in its own package?

package.json
```diff
@@ -82,6 +82,9 @@
   },
   "dependencies": {
     "classnames": "^2.2.6",
+    "chart.js": "^2.7.2",
```
oliverviljamaa: Even if we had the component here, we should not be bundling these dependencies.

matallo:
makes sense, happy to move it to its own repo, it started just as a line graph but then as we're only use it for rates changed the scope

will get rid of this as soon as I create the new repo, thanks for the suggestion!

PR #67, opened 2018-08-28
Modal

rs6g10:
# Modal Component

Adding a react modal component that uses bootstrap styles and animation.
It handles the following:
- Escape to close modal
- Modal with standard sizes.
- Fullscreen on mobile. This work is discussed with Mike Marter (@pixeljunkies ) who has confirmed the design for modals going full screen on mobile. It is recommended to use only one button in Mobile footer since it takes full width on mobile. If you need multiple buttons, you can stack them on each other, take guidance from your fellow designer. The styles will be moved to Transferwise bootstrap styles so we include no CSS in this project for modal.

### How does it look like?
modal-desktop
modal-mobile

This work is inspired by React bootstrap components.


src/index.js
```diff
@@ -7,3 +7,4 @@ export { default as Tooltip } from './tooltip';
 export { default as Stepper } from './stepper';
 export { default as FlowNavigation } from './flowNavigation';
 export { default as Popover } from './popover';
+export { Modal, ModalHeader, ModalBody, ModalFooter } from './modal';
```
oliverviljamaa: Can you explain why you chose the multiple components approach rather than the props approach that we use for other components (`Tooltip`, `Popover`, f.e.)?

src/modal/Animate.spec.js
```diff
@@ -0,0 +1,178 @@
...
+        <button className="trigger" onClick={this.toggle}>
+          Toggle
+        </button>
+        <TransitionGroup component="div">{this.state.showItem ? children : null}</TransitionGroup>
+      </div>
+    );
+  }
+}
+
+describe('Animate', () => {
```
oliverviljamaa: Can you explain why exactly we need this class? For the correct delay before adding or removing a class?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
```
oliverviljamaa: Just a tiny thing, this repo default imports `Types` ☺️

src/modal/index.css
```diff
@@ -0,0 +1,54 @@
+/*
```
oliverviljamaa: Why not move it to Bootstrap now if we want consistency?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+    super(props);
+    this.modalElement = null;
+    this.modalOriginalBodyPadding = null;
+
+    this.onClosed = this.onClosed.bind(this);
+    this.handleBackdropMouseDown = this.handleBackdropMouseDown.bind(this);
+    this.handleEscape = this.handleEscape.bind(this);
+
+    this.state = {
+      isOpen: props.isOpen,
```
oliverviljamaa: Controversial, I know, but we usually name our boolean flags without the `it` (see `inverse` elsewhere in this repo, f.e.).

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+          {Backdrop}
+        </Fragment>
+      );
+    }
+
+    return null;
+  }
+}
+
+Modal.propTypes = {
```
oliverviljamaa: Could we have these as static properties?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+      this.init();
+    }
+  }
+
+  componentDidMount() {
+    if (this.state.isOpen) {
+      this.setFocus();
+    }
+
+    this.isModalMounted = true;
```
oliverviljamaa: `this.isMounted`? As we know we're in the modal context.

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  static getDerivedStateFromProps = (nextProps, prevState) => ({
+    isOpen: nextProps.isOpen && !prevState.isOpen ? nextProps.isOpen : prevState.isOpen,
```
oliverviljamaa: Isn't this the same as `isOpen: nextProps.isOpen && !prevState.isOpen`? Also, why exactly do we need the previous state to be closed to open it for new props?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+      this.modalDialog &&
+      this.modalDialog.parentNode &&
+      typeof this.modalDialog.parentNode.focus === 'function'
+    ) {
+      this.modalDialog.parentNode.focus();
+    }
+  }
+
+  handleBackdropMouseDown(e) {
+    this.mouseElem = e.target;
```
oliverviljamaa: Exiting on click doesn't seem to work in the demo.

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  handleEscape(e) {
+    if (this.props.isOpen && e.keyCode === 27 && this.props.toggle) {
```
oliverviljamaa: Could we do something like `event.key === 'Escape'`?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+      this.modalTrigger = null;
+    }
+    this.modalElement = document.createElement('div');
+    this.modalElement.style.position = 'relative';
+    this.modalElement.style.zIndex = this.props.zIndex;
+    this.modalOriginalBodyPadding = getOriginalBodyPadding();
+
+    conditionallyUpdateScrollbar();
+
+    document.body.appendChild(this.modalElement);
```
oliverviljamaa: Let's use portals instead.

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+    if (this.modalTrigger) {
+      this.modalTrigger.focus();
+      this.modalTrigger = null;
+    }
+
+    if (this.bodyClassAdded) {
+      const modalOpenClassName = 'modal-open';
+      // Use regex to prevent matching `modal-open` as part of a different class, e.g. `my-modal-opened`
+      const modalOpenClassNameRegex = new RegExp(`(^| )${modalOpenClassName}( |$)`);
+      document.body.className = document.body.className
```
oliverviljamaa: With portals, we can get rid of this as well ☺️

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    return null;
+  }
+}
+
+Modal.propTypes = {
+  isOpen: PropTypes.bool,
+  size: PropTypes.oneOf(['lg', 'md', 'sm']),
+  toggle: PropTypes.func,
```
oliverviljamaa: Can you explain why we need this passed as a prop? Couldn't `Modal` know about its internals itself?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+  children: PropTypes.node,
+  fade: PropTypes.bool,
```
oliverviljamaa: What shall we change depending on that?

src/modal/Modal.js
```diff
@@ -0,0 +1,231 @@
...
+  zIndex: PropTypes.number,
```
oliverviljamaa: Is this necessary?

docs/ModalDocs.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+
+import { Modal, ModalHeader, ModalBody, ModalFooter, Select, Checkbox } from '../src';
+
+class ModalDocs extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      modal: false,
```
Tankenstein: `modalOpen` perhaps?

src/index.js
Tankenstein: Agree with this - it's not quite consistent

src/modal/Animate.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    ).toBe(0);
+  });
+
+  it('should transition classes from "fade" to "fade in" on enter', () => {
+    const onEnter = jest.fn();
+    const onExit = jest.fn();
+    const isOpen = false;
+    const wrapper = mount(
+      <Helper showItem={isOpen}>
+        <Animate onEnter={onEnter} onExit={onExit} key={Math.random()}>
```
Tankenstein: Just add some static keys: otherwise the test is not deterministic :(

src/modal/Animate.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+  afterEach(() => {
+    jest.clearAllTimers();
+  });
+
+  it('should transition classes from "fade" to "fade in" on appear', () => {
+    const isOpen = true;
+    const wrapper = mount(
+      <Helper showItem={isOpen}>
+        <Animate key={Math.random()}>Tere</Animate>
```
Tankenstein: Just add some static keys: otherwise the test is not deterministic :(

src/modal/Modal.js
Tankenstein: Oliver means without the `is`

src/modal/Modal.js
Tankenstein: ☝️

src/modal/Animate.js
```diff
@@ -0,0 +1,59 @@
...
+    className,
+    children,
+    innerRef,
+    ...otherProps
+  } = props;
+  const transitionProps = mergeProps(otherProps, TransitionPropTypeKeys);
+  const childProps = unMergeProps(otherProps, TransitionPropTypeKeys);
+
+  return (
+    <Transition {...transitionProps}>
```
Tankenstein: We can get rid of this entire class if we just use CSSTransition from the same library.

https://reactcommunity.org/react-transition-group/css-transition

src/modal/Animate.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    expect(
+      wrapper
+        .update()
+        .find('div.fade.in')
+        .hostNodes().length,
+    ).toBe(0);
+
+    wrapper
+      .find('.trigger')
+      .hostNodes()
```
Tankenstein: I think we can extract a lot of these bits here to tiny functions to make the test more readable

src/modal/Animate.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    );
+
+    expect(
+      wrapper
+        .update()
+        .find('div.fade')
+        .hostNodes().length,
+    ).toBe(0);
+    expect(
+      wrapper
```
Tankenstein: use `.find(...).exists()` instead of checking length

src/index.js
rs6g10: I do not understand the inconsistency here. This is not comparable with `Tooltip` or `Popover`. These are individual components that are required by users to use Modals. They all sit within the `modal` folder and exporting them should be as simple as how it is done. Are you suggesting `ModalTitle`, `ModalBody`, `ModalFooter` have their own folders? If yes, then I can extract them out in different folders and then reference them as per your suggestion.

src/modal/Animate.spec.js
rs6g10: Overall, yes.

src/modal/index.css
rs6g10: Because, we we move to Bootstrap staight-away, people are going to get surprises everywhere, since it's referenced in many places. We will roll it out with react-components, we will have an example to show to everyone and then roll out the changes upon agreement.

src/modal/Modal.js
rs6g10: I was told by our team that we use `isOpen`, `isClosed` type boolean naming convention. Am I missing something here?

src/modal/Modal.js
rs6g10: Why static properties?

src/modal/Modal.js
rs6g10: I'll change it 👍

src/modal/Modal.js
rs6g10: I have removed that functionality, because often quite a lot when we have forms in modals, users accidentally click outside to close the form. This came out as a requirement for our team.

src/modal/Modal.js
rs6g10: I thought about it and so I aptly named the method that it handlesEscape, so I didn't have to create a variable escape that has value 27 :)

src/modal/Modal.js
rs6g10: I see what you're saying. I'll see if we could do it, though I don't see this as an anti-pattern in anyway.

src/modal/Modal.js
rs6g10: 👍

src/modal/Modal.js
rs6g10: What `prop` are we talking about? `toggle`?

src/modal/Modal.js
rs6g10: ???

src/modal/Modal.js
rs6g10: When you use cssmin in future, you're going to have a good time realising how zIndex changes 👍 . I've added this as part of the experience I've had with [cssnano](https://cssnano.co/).

src/modal/Modal.js
rs6g10: I'll simplify this :)

src/modal/Modal.js
rs6g10: Okay, I've always used `PropTypes`, I'll change it to use `Types` :)

src/modal/Animate.spec.js
rs6g10: 👍

src/modal/Modal.spec.js
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+import React from 'react';
+import { mount, shallow } from 'enzyme';
+import Modal from './Modal';
+import ModalHeader from './ModalHeader';
```
mstaicu: I don't think you need these imports here, the header, body and footer

src/modal/Modal.js
oliverviljamaa: Yes.

src/modal/Modal.js
oliverviljamaa: But what's the benefit of even having the value 27?

src/modal/Modal.js
oliverviljamaa: Are we... sure about this...? I was very surprised to not see this work as it's my default way of closing a modal.

src/modal/Modal.js
oliverviljamaa: As that's usually the pattern for stateful components (keeping the API of the component on the top, as it's important).

src/modal/Modal.js
oliverviljamaa: `components` seem to use boolean flags without the `is`, or are there counter-examples?

src/modal/index.css
oliverviljamaa: Thanks, sounds good! How will we ensure this code is not here in, say, 3 months?

src/modal/Modal.js
oliverviljamaa: (I've personally always used `PropTypes` as well 😄)

src/index.js
oliverviljamaa: I believe we're talking about different things – we're talking about not having all these components, but just the `Modal` component with props `header`, `body`, and `footer`, similar to `Popover` (in this same repo).

src/modal/Modal.js
oliverviljamaa: Why would we ever not `fade`?

src/modal/Modal.js
oliverviljamaa: Can you elaborate, please? Does the plugin modify certain `z-index`? If yes, is it not possible to disable it?

src/modal/Modal.js
oliverviljamaa: I agree, this here would be fine, but the `className` regex is the main reason.

src/index.js
Tankenstein: Where "body" would prolly be children of the modal.

src/modal/Modal.js
Tankenstein: I would remove the fade property at all. Quite safe to say we'll always fade.

src/modal/Modal.js
Tankenstein: I do see this as an antipattern. You're having to manage dom and dom lifecycles manually. React  (and all similar frameworks) encourage you to always leave dom management to the framework, except where it's impossible. In this case, it would be more readable and actually much simpler of an implementation if you just used portals, plus you wouldn't have to manage the dom yourself, which is tricky to get absolutely right, and more complicated to test.

src/modal/Modal.spec.js
Tankenstein: Something's wrong with the eslint disable rule you put up there, as this should not have passed tests. I think you accidentally made eslint ignore this file entirely :)

src/modal/Modal.js
Tankenstein: I think this is something we can add later on if we need it. I'd remove it until we need it.

src/modal/Modal.js
rs6g10: So all I'm doing here is adding a class that freezes scrolling on body and it removes this class on  destroy. Can you please be a bit more elaborative on how React Portal can help me with this better than how it is done now? An example would be best :)

src/modal/Modal.js
rs6g10: I'll change 27 to ESC_KEY variable if that helps :)

src/modal/Modal.js
rs6g10: Yes! It was specifically requested for, you can see it in commit history that I was handling it before (as we discussed this in our initial meeting :) )

src/modal/Modal.js
rs6g10: Can you provide me with an example or some documentation for this please?

src/index.js
rs6g10: This then won't act as open as a component it will become. I believe, if we only ask for title, content and buttons (for both headers and footers); then we'll be taking away quite some big functionality of modals. Also, this is very much similar to how bootstrap works today. If we are aiming for a very custom modals, then I believe we will lose quite some users of this component. This modal came out as requirement that requires all these components. What do you think?

src/modal/Animate.spec.js
rs6g10: I am sorry, I don't understand. Can you please be a bit more elaborative about this?

src/modal/Animate.js
rs6g10: Can you provide an example in our context? How will it help instead of what we currently have from the same library?

src/modal/Modal.js
oliverviljamaa: I personally would just replace this:
```js
if (this.props.isOpen && e.keyCode === 27 && this.props.toggle)
```
with this:
```js
if (this.props.isOpen && event.key === 'Escape' && this.props.toggle)
```
(and rename `e` to `event`)

oliverviljamaa: I may be missing something on browser support here.

src/index.js
oliverviljamaa: `This then won't act as open as a component it will become.` I don't think I understand. If this `Modal` component has, say, 4 props: `open`, `header`, `body`, and `footer`, what functionality are we taking away? (I may be completely misunderstanding here, sorry!)

src/modal/Modal.js
oliverviljamaa: Of course!
From React docs:
https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values
From our usage in this repo:
https://github.com/transferwise/components/blob/master/src/moneyInput/MoneyInput.js#L19

src/modal/Modal.js
oliverviljamaa: I personally don't mind that much, as long as it's agreed on with the design team!

src/modal/Modal.js
rs6g10: Got it - angular world `is` type booleans. Components world - `normal` booleans :)

src/index.js
rs6g10: If we component out ModalHeader, ModalBody and ModalFooter, we can make sure it lies in line with our style-guide. If we don't and we only expect users to provide their own header, body and footer - then we let users take control over how they use our modals. What do you think?

oliverviljamaa: I personally cannot see the difference – in both ways, we can put anything in the `.modal-header`, `.modal-body`, and `.modal-footer` without having to worry about the Bootstrap-specific classes. Can you explain how users can take more control?

rs6g10: This you can do, without using `ModalHeader`, `ModalBody` & `ModalFooter` anyways. When I say, if users provide their own header, body and footer (which they still can), we are letting them take control over how they'd want to style the modal content, thus taking control over what's inside the modal. Are you suggesting the new 3 props i.e. `header`, `body`, `footer` to be required or optional? Because if they're made optional then users would then still need to create their own `header`, `body`, `footer` and they can still use our components or use their own with whatever styling they want. Does this make sense? I think this conversation is getting a bit confusing now :/

Tankenstein: @rs6g10 Actually, we're trying to *avoid* giving that flexibility. A big reason why we have these common component repos is to enforce styling.

src/modal/Modal.js
Tankenstein: You're manually handling the lifecycle of the modal element here, and portals help you manage that automatically. It even has a modal example right there in the docs: https://reactjs.org/docs/portals.html it's the idiomatic way to implement stuff like modals in react.

src/modal/Modal.js
Tankenstein: We require a compilation step anyways, we're using jsx etc 😄

PR #73, opened 2018-09-17
Make money input not have decimals for integer amounts

Tankenstein:
Changes formatting of money input so that when no decimals are provided, it shows an integer amount when formatting. If a user inputs decimals, we format with the currency's decimals.


src/moneyInput/currencyFormatting.js
```diff
@@ -42,8 +42,22 @@ function getCurrencyDecimals(currency = '') {
   return DEFAULT_CURRENCY_DECIMALS;
 }
 
+function countDecimalPlaces(number, locale) {
```
oliverviljamaa: What stops us from using `@transferwise/formatting` logic?

Tankenstein: We need parsing here, which formatting won't support anyhow. Felt weird to include that but then also write the parsing stuff, not sure, wdyt?

oliverviljamaa: I'd prefer that, as at the moment we have the exact same logic in two places (currency decimals config object, number locale support check, and the formatting itself). For example, if we used `formatCurrency` from `@transferwise/formatting` already, we wouldn't have had to repeat the logic needed for this PR's change, as it already exists for `formatAmount`.

Doesn't seem that bad to:
1) Parse a string to a number
2) `formatAmount` that number
3) Return the formatted number

PR #74, opened 2018-09-24
Change Select component's styles

Tankenstein:
Currency options now show the flag in every device size and keyboard focused options have the same style as hovered elements, rather than active ones.


src/select/Select.spec.js
```diff
@@ -298,18 +310,18 @@ describe('Select', () => {
...
-    expect(findNthListElement(3).hasClass('active')).toBe(false);
+    expect(findNthListElement(3).hasClass('tw-dropdown-item--focused')).toBe(false);
     doTimes(3, () => component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.DOWN)));
-    expect(findNthListElement(3).hasClass('active')).toBe(true); // skips header!
+    expect(findNthListElement(3).hasClass('tw-dropdown-item--focused')).toBe(true); // skips header!
 
     doTimes(4, () => component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.DOWN)));
-    expect(findNthListElement(3).hasClass('active')).toBe(false);
-    expect(findNthListElement(8).hasClass('active')).toBe(true); // skips header and separator again!
+    expect(findNthListElement(3).hasClass('tw-dropdown-item--focused')).toBe(false);
```
oliverviljamaa: Could we encapsulate this to not repeat it?
`expect(focusIsOnListElementN(3)).toBe(false);`

Tankenstein: Ya sure 👍

src/moneyInput/MoneyInput.js
```diff
@@ -166,7 +166,7 @@ export default class MoneyInput extends Component {
               <Fragment>
                 <i className="tw-money-input__keyline" />
                 <i
-                  className={`currency-flag currency-flag-${selectedCurrency.currency.toLowerCase()} hidden-xs m-r-2`}
+                  className={`currency-flag currency-flag-${selectedCurrency.currency.toLowerCase()} m-r-2`}
```
oliverviljamaa: Don't we still want to hide the selected currency flag?

oliverviljamaa: Or is this something else?

Tankenstein: Originally, yes. However, i took a look. The smallest size we need to deal with is the SE, on which this starts to get nasty around 10 mil, so i figured let's just always show it

Tankenstein: Maybe the label gets nasty actually?

oliverviljamaa: The label was the original reason, yeah, I'd keep the behaviour the same for now and keep hiding?

Tankenstein: Ok 👍

src/select/option/Option.spec.js
```diff
@@ -35,4 +35,12 @@ describe('option', () => {
     component.setProps({ currency: 'hustle' });
     expect(component.find('i.currency-flag.currency-flag-hustle').length).toBe(1);
   });
+
+  it('does not show currency icons on mobile if it is selected', () => {
```
oliverviljamaa: Wait, I'm confused here... Isn't this a dropdown option and we would never hide its flag?

Tankenstein: This is also used for the "button" internal view of the select component

Tankenstein: Check out the `renderButtonInternals` method on `Select`

oliverviljamaa: 😮

PR #77, opened 2018-10-04
Proxy over other properties of modal to the rendered component

Tankenstein:
When the `Modal` component gets passed any unknown properties, it'll proxy them to the modal instance that gets rendered.


coolsebz:
writing here for posterity even though we talked through it, it would be useful to document `className` as the official way to pass styles downstream

other than that, looks good to me 👍

src/modal/ModalDialog/ModalDialog.js
```diff
@@ -38,14 +38,15 @@ export default class ModalDialog extends Component {
 
     return (
       <div
-        className="modal fade"
+        className={`modal fade ${className}`}
```
oliverviljamaa: Why not use `classNames`?

Tankenstein: i actually did and removed, less readable/understandable immediately in this case

`className={`modal fade ${className}`}`

vs

`className={classNames('modal fade', { [className]: className })}`

wdyt?

coolsebz: think that would be more confusing, since the normal property is `className` anyway, that's what i expect most people to reach for

oliverviljamaa: `className={classNames('modal fade', className)}` should work imo.

oliverviljamaa: But not necessarily better.

Tankenstein: for that simple of a case string interp is fine, even better imo

coolsebz: wdym should work, isn't that how you normally use classNames? 😁

Tankenstein: @coolsebz that was a reply to my comment, where i did the object key version

PR #78, opened 2018-10-05
Allow overriding of class names for currency flags in select component

phildobsontw:


src/select/option/Option.spec.js
```diff
@@ -43,4 +43,15 @@ describe('option', () => {
     component.setProps({ selected: true });
     expect(flagHiddenOnMobile()).toBe(true);
   });
+
+  it('Can override class names to contain a localIdentName', () => {
```
Tankenstein: a) `Can` -> `can`
b) Drop the 'to contain a localIdentName' to just leave 'can override class names' or something

Tankenstein:
In order to release, you need to increment package.json and package-lock.json versions and add a changelog entry :)

phildobsontw:
👍Thanks for reviewing so quickly! Comments addressed, should be good to go now?

PR #79, opened 2018-10-08
Make modals not destroy ssr

Tankenstein:
Previously, modals would throw when trying to ssr as they are accessing document directly, this release fixes that issue.


src/modal/Modal.js
```diff
@@ -23,4 +23,11 @@ Modal.propTypes = { open: Types.bool, ...dialogPropTypes };
 Modal.defaultProps = { open: false, ...dialogDefaultProps };
 
 export { Modal };
-export default props => createPortal(<Modal {...props} />, document.body);
+export default props => {
+  const mountPoint = getMountPoint();
+  return mountPoint ? createPortal(<Modal {...props} />, mountPoint) : '';
```
rs6g10: nitpick but use `&&` instead of empty string.

Tankenstein: I'd rather empty string than undefined when someone tries to render it on the server

coolsebz: think it may be worth noting that if a user tries to render a Modal on the server for a specific path (as an example) this won't work, right?

Tankenstein: Yep. I'll add a note

Tankenstein: Once we start to need it let's use the universal portals package, right now they're not recommending to use it as it has a biggie issue

PR #80, opened 2018-10-11
Make class name overriding consistent across select component

phildobsontw:


src/select/Select.js
```diff
@@ -229,16 +229,16 @@ export default class Select extends Component {
   renderSearchBox() {
     const { searchValue, searchPlaceholder } = this.props;
     return (
-      <li className="tw-dropdown-item--divider">
+      <li className={this.style('tw-dropdown-item--divider')}>
```
oliverviljamaa: Can you explain why we've gone with the `classNames` object as a prop? To me, it seems easier to just have `className` as a string.

oliverviljamaa: What are the reasons for needing to add classes for component internals?

phildobsontw: Sorry if i've misunderstood the question - this instance has a single class that needs overriding, so I chose `{this.style('tw-dropdown-item--divider')}` rather than using template literals ```{`${this.style('tw-dropdown-item--divider')}`}``` because I thought it read slightly easier (less curly braces). If there's a performance difference or any other reason - happy to change this!

phildobsontw: 2) Just for context - we're seeing styling issues in the comparison component, e.g:

image

Taking the flag misalignment for example, this is happening because the CSS highlighted below isn't selected due to the mix of namespaced + non namespaced class names.
https://github.com/transferwise/components/blob/93e40815216eb7c7a018a3983a6ab864e927dd91/src/select/Select.less#L19

Namespacing both fixes this.

src/select/Select.js
```diff
@@ -260,7 +260,10 @@ export default class Select extends Component {
       <li
         onClick={this.createSelectHandlerForOption({ placeholder })}
         onKeyPress={this.createSelectHandlerForOption({ placeholder })}
-        className="tw-dropdown-item--clickable tw-dropdown-item--divider"
+        className={classNames(
```
phildobsontw: FYI - when using string template literals here, prettier adds a trailing comma to the second which I thought looked a bit strange, so I used classNames instead. Here's how it would look if I were to use string templates after prettify has run in the pre-commit hook:
```className={`${this.style('tw-dropdown-item--clickable')} ${this.style(
          'tw-dropdown-item--divider',
        )}`}```.

If we're happy with the trailing comma, happy to change this to use a template literal.

src/select/Select.js
oliverviljamaa: Thanks for the thorough explanation! However, it's still not completely clear to me, sorry – why does the `margin-right` not apply?

phildobsontw: The problem basically lies in that in comparison components we add namespaces to all of the CSS class names so that we’re isolated from any changes in whichever app is hosting us. This means that in our build process the css in the file below is only applied when `tw-dropdown-item_{namespace}` and `currency-flag_{namespace}` is applied. The classes that are actually getting applied on the element are a mismatch of namespaced and non-namespaced however, in this example `tw-dropdown-item` and `currency-flag_{namespace}` which means the css isn’t applied.

PR #81, opened 2018-10-26
hide close button on FlowNavigation if onClose is undefined

Kaarel94:
This is useful for flows (like refresh cycle) where we need to force the user to go through the flow and in those cases showing a close button that doesn't do anything is unwanted.


src/flowNavigation/FlowNavigation.js
```diff
@@ -22,11 +22,15 @@ const FlowNavigation = ({ steps, activeStep, avatarUrl, done, profileType, onClo
         </div>
         <div className="col-lg-2 col-xs-6 col-lg-push-8 text-xs-right m-lg-t-1">
           {done ? '' : <Avatar url={avatarUrl} profileType={profileType} />}
-          <button
-            className={`btn-unstyled tw-flow-navigation__close-button
+          {onClose ? (
```
mikknarusk: `onClose && (<button />)`?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v4.5.4
+## Close button for FlowNavigation is now hidden if onClose is undefined
```
Tankenstein: I think this should be a minor version bump, not patch, as you're adding functionality, imo :)

PR #82, opened 2018-10-30
Add RadioOption, CheckboxOption, and NavigationOption

oliverviljamaa:
# v4.7.0
## Add RadioOption, CheckboxOption, and NavigationOption

image


docs/NavigationOptionDocs.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable react/prefer-stateless-function */
+
+import React, { Component } from 'react';
+import '@transferwise/icons/dist/icons.min.css'; // eslint-disable-line
+
+import OptionDocs from './common/OptionDocs';
+import { NavigationOption } from '../src';
+
+class NavigationOptionDocs extends Component {
```
Tankenstein: Why not a functional component?

docs/RadioOptionDocs.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable */
```
Tankenstein: Why the global disables?

src/common/Option/Option.js
```diff
@@ -0,0 +1,58 @@
...
+
+const Option = ({
+  icon,
+  title,
+  description,
+  onClick,
+  htmlFor,
+  disabled,
+  typeSpecificContent,
+  complex,
```
Tankenstein: I'm not sure about the `complex` prop naming. In addition, `typeSpecificContent` could prolly be simplified to `button` or `action`?

oliverviljamaa:
Will work on keyboard support for radio option and navigation option.

OFFLlNE:
Could you post a screenshot how it would look like without the subtext (just `XXXXXXXXX option`)

src/common/CheckboxButton/CheckboxButton.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+const CheckboxButton = ({ checked, onClick, disabled, className }) => (
```
Tankenstein: Honestly: i'd remove the current checkbox/radio components as they're shite. Replace the current ones with these, maybe, make a breaking release?

src/common/Option/Option.js
oliverviljamaa: Probably `button` may make sense. Or maybe `trigger`? Which would you prefer?

docs/RadioOptionDocs.js
oliverviljamaa: The ESLint loader is annoying for development, forgot those in, thanks!

src/common/CheckboxButton/CheckboxButton.js
oliverviljamaa: Can you explain what you mean? As this is just the `button` or `trigger` part, for us to be able to reuse it in the button-styled and option-styled components.

oliverviljamaa:
@OFFLlNE 
image

src/common/Option/Option.js
Tankenstein: button i think

docs/CheckboxOptionDocs.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import '@transferwise/icons/dist/icons.min.css'; // eslint-disable-line
```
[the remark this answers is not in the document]
oliverviljamaa: It's tougher with Angular, for React, I expose React components for every icon, so they can be pulled in on-demand.

docs/common/OptionDocs.js
```diff
@@ -0,0 +1,135 @@
...
+            </div>
+            <div className="col-md-6">
+              <Option
+                id={`${type}-option`}
+                name={`${type}-option`}
+                title={title}
+                description={getDescriptionForComplexState(complex)}
+                complex={complex}
+                disabled={disabled}
+                icon={<FastFlagIcon />}
```
Tankenstein: Instead of calling that prop `Icon`, i would maybe make it more generic, like `media`, as you can have avatars or such in here.

src/common/Option/Option.js
oliverviljamaa: Changed to `button`.

docs/common/OptionDocs.js
oliverviljamaa: Changed.

Tankenstein:
One nitpick: perhaps focus styles on these should be the same as hover styles?

When i use my keyboard to use the navigation option, i can very barely see the arrow getting focused

oliverviljamaa:
@Tankenstein Yup, these changes would be in Bootstrap, though.

Tankenstein:
Fair point

src/radio/Radio.js
```diff
@@ -1,39 +1,22 @@
...
-          <span className="tw-radio-check" />
-        </button>
-        {label}
-      </label>
-    </div>
-  );
-};
+import RadioButton from '../common/RadioButton';
+
+const Radio = ({ label, disabled, id, name, checked, onChange }) => (
```
williambaumgartner: Why not rest spread all the props? Is it for clarity?

src/checkbox/Checkbox.spec.js
```diff
@@ -15,65 +16,87 @@ describe('Checkbox', () => {
...
+    expect(checkboxButton().prop('checked')).toBe(true);
   });
+
+  it('passes disabled to checkbox button', () => {
+    expect(checkboxButton().prop('disabled')).toBe(false);
+    component.setProps({ disabled: true });
+    expect(checkboxButton().prop('disabled')).toBe(true);
+  });
+
+  function isDisabled() {
```
williambaumgartner: I don't like the way we rely on *scope* for these **selector functions**.

```javascript
// i feel pure functions are much better & concise
// for ex.
const input = component => component.find('input');
const id = component => input(component).prop('id');
```

src/common/Option/Option.js
```diff
@@ -0,0 +1,48 @@
...
+
+import './Option.less';
+
+const Option = ({ media, title, description, onClick, htmlFor, disabled, button, complex }) => (
+  <label
+    className={classNames('media decision', { 'decision-complex': complex, disabled })}
+    onClick={event => {
+      if (onClick) {
+        event.preventDefault();
+        onClick();
```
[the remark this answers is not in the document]
oliverviljamaa: When would we use it? I find it cleaner to not bubble native element events up from our components.

src/radio/Radio.js
oliverviljamaa: To be explicit, yes, but I'm happy to change if you think it's better.

src/checkbox/Checkbox.spec.js
oliverviljamaa: Regarding functional purity, we have opted for relying on the scope in tests (and tests only, of course) for readability:
```js
expect(checkboxButton().prop('disabled')).toBe(true);
```
vs
```js
expect(checkboxButton(component).prop('disabled')).toBe(true);
```

Regarding using arrow functions, I do think they're nicer to read, but we would have to either rely on the implicit fact that `it`s get executed after the content of the parent function. I'm kind of torn here, as the arrow functions are indeed more readable.

WDYT? @williambaumgartner

src/radio/Radio.js
williambaumgartner: I don't have a strong opinion on it but I find `propTypes` to be really good in that regards.

src/common/Option/Option.js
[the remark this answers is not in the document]
williambaumgartner: Well, it could be used to stop the propagation up... but I'm wondering if we have use cases in which this would be necessary.

src/checkbox/Checkbox.spec.js
williambaumgartner: Oh, I did not know that, that's ok.

Maybe **arrow** functions then? I feel they're easier to read, unless we need to mess with **`this`** binding for some reason.

They'd have to be moved up to the top though.

src/common/Option/Option.js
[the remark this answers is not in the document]
oliverviljamaa: In terms of universal components, I'd prefer to be restrictive and loosen up when there's a clear use case.

src/checkbox/Checkbox.spec.js
oliverviljamaa: @williambaumgartner The arrow functions work even when they're in the bottom (as the `it` functions' second arguments are not immediately invoked). Moving them to the top will move low-level details to the most noticeable place, which is not desired, but having them in the bottom and working feels a bit like magic.

I personally would be willing to accept that magic for readability... How about you?

src/common/CheckboxButton/CheckboxButton.js
oliverviljamaa: Discussed separately, will keep the breaking change to a new version.

src/radio/Radio.js
oliverviljamaa: Done!

src/checkbox/Checkbox.spec.js
oliverviljamaa: @williambaumgartner Refactored to arrow functions!

docs/common/OptionDocs.js
```diff
@@ -0,0 +1,135 @@
...
+  return complex
+    ? 'Complex items should be used for items with long descriptions. Their button and media will be aligned to the top.'
+    : 'Normally, the button and icon are vertically centered.';
+}
+
+function boolProp(name, value) {
+  return value
+    ? `
+  ${name}`
+    : '';
```
dozortsev: a bit weird code formatting here :) is it a Prettier?

docs/CheckboxOptionDocs.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+
+import OptionDocs from './common/OptionDocs';
+import { CheckboxOption } from '../src';
+
+class CheckboxOptionDocs extends Component {
```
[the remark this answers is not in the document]
oliverviljamaa: Good comment. Yes, we have, and I believe everyone agrees it'd be better, but no one has made the leap yet. @rs6g10 and anyone else who's keen, wanna hack away at it sometime this week?

docs/Docs.less
```diff
@@ -1,5 +1,6 @@
 @import '~bootstrap/dist/css/bootstrap.css';
 @import '~currency-flags/dist/currency-flags.min.css';
+@import '~@transferwise/icons/dist/icons.min.css';
```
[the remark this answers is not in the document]
oliverviljamaa: The icons will be provided by the consumer of the component, yes, here we're just including them for the docs.

docs/common/OptionDocs.js
oliverviljamaa: Prettier and the fact that we need the newline before the `${name}` 🙂

src/navigationOption/NavigationOption.less
```diff
@@ -0,0 +1,22 @@
+@import '~bootstrap/less/variables';
+
+.tw-navigation-option__chevron {
```
lapostoj: Add the same `// TODO: Move to Bootstrap if considered reasonable`

src/radio/Radio.js
```diff
@@ -1,39 +1,22 @@
...
-        {label}
-      </label>
-    </div>
-  );
-};
+import RadioButton from '../common/RadioButton';
+
+const Radio = ({ label, ...otherProps }) => (
+  <div className="radio">
+    {/* eslint-disable jsx-a11y/label-has-for */}
```
lapostoj: Maybe we could just put the "for" label to not need the disabling? Though it means we would need to pass the id to the RadioButton I guess

src/common/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,48 @@
...
+
+import './RadioButton.less';
+
+const RadioButton = ({ id, name, checked, onChange, disabled }) => (
+  <Fragment>
+    <input
+      type="radio"
+      className="sr-only"
+      id={id}
+      value={id}
```
lapostoj: `value={id}`? 🤔

src/common/CheckboxButton/CheckboxButton.spec.js
```diff
@@ -0,0 +1,42 @@
...
+
+import CheckboxButton from '.';
+
+describe('Checkbox button', () => {
+  let component;
+  beforeEach(() => {
+    component = shallow(<CheckboxButton onClick={jest.fn()} />);
+  });
+
+  it('has checked class when checked', () => {
```
StevePole: How do feel about structuring these in a more BDD style?  So, use additional describe blocks to explain the scenario, and 'it' blocks to explain the expectation.  In this case you would have:
```
describe('Given a checkbox, when initialised', () => {
  let component;
  beforeEach(() => {
      component = shallow(<CheckboxButton onClick={jest.fn()} />);
  });
  it('should not have checked class', () => {
    expect(hasCheckedClass()).toBe(false);
  });
  it('should not have aria-pressed attribute', () => {
    expect(hasAriaPressedAsTrue()).toBe(false);
  });
  it('should not call click handler', () => {
    expect(onClick).not.toBeCalled();
  });
  describe('when prop is checked', () => {
    beforeEach(() => {
      component.setProps({ checked: true });
    });
    it('should have checked class', () => {
      expect(hasCheckedClass()).toBe(true);
    });
    it('should have aria-pressed attribute', () => {
      expect(hasAriaPressedAsTrue()).toBe(true);
    });
  }
  describe('when clicked', () => {
    beforeEach(() => {
      const onClick = jest.fn();
      component.setProps({ onClick });
    });
    it('should call click handler', () => {
      expect(onClick).toBeCalled();
    });
  });
});
```

I find these tests readable as is, but as tests get more complex I've found this to scale better.  The main reason is that test setup is not interspersed with assertions, and test setup does not need to be repeated.  For example, you set `checked: true` in a couple places, this only happens once in this format.  You do add more 'documentation' code via the describe blocks, but I don't think that's a bas thing.  

However, my main interest in pushing this approach is that I think it will help folks to write better tests, as it forces you to think about what behaviours you have and what should happen as a result.

docs/Docs.less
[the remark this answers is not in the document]
lapostoj: You can remove them from "peerDependencies" if they are used only for docs though!

src/common/RadioButton/RadioButton.js
oliverviljamaa: Yes, the `value` prop is necessary for the user agent to differentiate between different radios with the same `name` and necessary for keyboard support, but as it has the same characteristics as the `id` (unique string), we can remove the burden from the consumer.

src/common/CheckboxButton/CheckboxButton.spec.js
oliverviljamaa: I agree, sometimes we use this structure, but it's never clear when we should. Here it would work well.

oliverviljamaa: Some questions, though – where would you include the assertion that the handler is not called before the click simulation?

oliverviljamaa: Made changes in 624f9c61b3524d96282cc288ace59ad0954ae6b3.

I did not include the `should`s, as we do not use them any more (no real value, just repetition) and I left `it`s that would be the only children of the `describe` be for readability. I also moved a bunch of assertions that would have a matching assertion in a `when X` in a `by default` `describe` to keep them on the same level.

WDYT?

src/common/CheckboxButton/CheckboxButton.js
```diff
@@ -0,0 +1,33 @@
...
+  <button
+    type="button"
+    className={classNames('tw-checkbox-button', { checked }, className)}
+    aria-pressed={checked}
+    onClick={() => {
```
rs6g10: Is there a reason why this can't be passed onto as just the prop? Why a function call here, given onClick already comes from the props?

src/checkboxOption/CheckboxOption.js
```diff
@@ -0,0 +1,51 @@
...
+  complex,
+  disabled,
+}) => {
+  const sharedProps = { media, title, description, name, complex, disabled };
+
+  return (
+    <Option
+      {...sharedProps}
+      onClick={() => {
+        if (!disabled) {
```
rs6g10: Can this not be handled on the component that calls this component? Why the logic based out on props?

src/common/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,48 @@
...
+        onChange(!checked);
+      }}
+      disabled={disabled}
+    />
+    <button
+      type="button"
+      className={classNames('tw-radio-button', { checked })}
+      disabled={disabled}
+      aria-pressed={checked}
+      tabIndex="-1"
```
rs6g10: `tabIndex="-1"` Is the idea to make this focusable after errors? Why handle it in the component? Is this testable?

src/checkboxOption/CheckboxOption.js
oliverviljamaa: Would you rather have the disabled check on the component consumer's side? If yes, what are the use cases for ever needing to call the change handler on clicking a disabled item?

src/common/CheckboxButton/CheckboxButton.js
oliverviljamaa: This is to rid the click handler from the `event` that is relevant to the underlying `button`, but not to the `CheckboxButton` itself.

src/common/RadioButton/RadioButton.js
oliverviljamaa: Nice catch. This is to not make it focusable, so we could have radio button behaviour.

src/checkboxOption/CheckboxOption.js
rs6g10: I don't know. It feels a bit weird to have this logic in `onClick`. The client should know about `onClick`, right? And it wouldn't be clickable while disabled anyways, right? But if you think this is the right case for it, then go for it.

src/common/CheckboxButton/CheckboxButton.js
rs6g10: It feels a bit odd to be honest, it's not clearly understandable why we're doing this - 
```
onClick={() => onClick();}, don't you think?
```

src/checkboxOption/CheckboxOption.js
oliverviljamaa: I'm not sure anymore, will take a look again to see whether I can handle this natively, thanks.

src/common/CheckboxButton/CheckboxButton.js
oliverviljamaa: We're not even doing this, as we're not returning the result of the `onClick`. We only care about calling the click handler _when_ the click is made, and don't need the `event`.

What would you suggest? (happy to discuss F2F as well to get a better understanding!)

rs6g10: Apologies for the delay. Okay, so then I would suggest something very simpler, like below:
screen shot 2018-11-07 at 10 03 27

oliverviljamaa: No problem at all, this is what I'd normally do if I ever felt the need to bubble the `event` up, but here, I don't think it's desired and will actually end up being incorrect information, wouldn't it?

rs6g10: cool

src/common/CheckboxButton/CheckboxButton.spec.js
StevePole: Not bothered about the 'shoulds', still a bit of an odd mix of ordering imo.  Lots of assertions that end with 'when' and 'on' rather than beginning with it.  Makes it a bit harder to read when sometimes the scenario is described before the expected behaviour and sometimes after.

e.g.
'is disabled when disabled flag is passed'
'calls click handler on click'

Even if keeping those description in the 'it', and avoiding additional describe blocks, I would prefer to preserve the 'given, when, then' order, so:

'when disabled flag is passed, is disabled'
'when clicked, calls click handler'.

But again, not a big deal.

williambaumgartner:
image

@OlaviOjamaa
If you refer to the image above. I might need a change. Either a way to pass in a `checked/selected` icon or have a `checked` state.

oliverviljamaa:
@williambaumgartner Yeah, this is part of _chapters_, usage of which I was not certain about. We could work together on adding this in a future release, WDYT?

williambaumgartner:
@oliverviljamaa Perfect. 👍

src/checkboxOption/CheckboxOption.js
oliverviljamaa: I tried doing it natively, but it wouldn't be possible for the navigation option, as there's no `input` that could be controlled by a `label`, there's just a `button`.

Do you see any use cases for calling the change handler on disabled item click?

PR #84, opened 2018-11-23
onClose prop not required for FlowNavigation

Kaarel94:
With #81 it's now a valid use case to have onClose be undefined for FlowNavigation, but forgot to change propTypes in that PR. This fixes that.


src/flowNavigation/FlowNavigation.js
```diff
@@ -48,6 +48,7 @@ FlowNavigation.defaultProps = {
   done: false,
   profileType: ProfileType.PERSONAL,
   onGoBack: null,
+  onClose: undefined,
```
oliverviljamaa: Could we default to `null` for consistency with `onGoBack` and to not _define something as undefined_.

PR #86, opened 2018-11-28
Initialized Telephone Component

AndreaTW:


docs/PhoneNumberInputDocs.js
```diff
@@ -0,0 +1,73 @@
...
+      required: false,
+      disabled: false,
+      value: '+44534353563',
+      locale: 'en-EN',
+    };
+  }
+
+  render() {
+    const docsCode = `<PhoneNumberInput
+  locale=${this.state.locale}
```
taaviaasver: Please match example code with other component examples:

- `locale=${this.state.locale}` -> `locale="${this.state.locale}"`
- `value=${this.state.value}` -> `value="${this.state.value}"`
- `disabled=${this.state.disabled}` -> `disabled={${this.state.disabled}}`
- `required=${this.state.required}` -> `required={${this.state.required}}`

docs/PhoneNumberInputDocs.js
```diff
@@ -0,0 +1,73 @@
...
+              {/* eslint-enable react/jsx-indent */}
+              <p>Every property that is false or undefined can be omitted.</p>
+            </div>
+
+            <div className="col-md-6">
+              <div className="row">
+                <div className="col-md-6">
+                  <Checkbox
+                    label="Required"
+                    checked={false}
```
taaviaasver: This should be `this.state.required` instead of hard-coded value

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
...
+class PhoneNumberInput extends PureComponent {
+  // PropTypes definitions.
+  static propTypes = {
+    required: Types.bool,
+    disabled: Types.bool,
+    value: Types.string,
+    onChange: Types.func.isRequired,
+    locale: Types.string,
+    searchPlaceholder: Types.string,
+    size: Types.oneOf(['sm', 'md', 'lg']),
```
taaviaasver: I'd like to see `size` and `placeholder` in example documentation as well.. or at least a mention that you can also pass in `size` and `placeholder`.

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  // Default Props.
+  static defaultProps = {
+    required: false,
+    disabled: false,
+    value: '',
+    locale: 'en_GB',
+    searchPlaceholder: 'Prefix',
```
taaviaasver: Users with locale `es` (for example) will see non translated value

If its implemented the same way in `styleguide-components` version then I'm fine with it..  just pointing it out that it would be a bit better if the placeholder would be translated.

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,4 @@
+
```
taaviaasver: I think it doesn't make sense to include this empty stylesheet

src/phoneNumberInput/utils/pnoneNumberUtils.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return undefined;
+};
+/**
+ * Given a valid locale it returns the correspondent prefix if found or +44 otherwise.
+ * @param {string} locale - a string that represent the locale ex:'es-ES'
+ * @returns {string}
+ */
+export const setDefaultPrefix = locale => {
+  const localeCountry = localeService.getCountryFromLocale(locale);
```
StevePole: Where does localeService come from?  As we're not using Angular DI, I would expect to see it as an import at the top of the file?

src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -0,0 +1,189 @@
...
+      select = wrapper.find(PREFIX_SELECT_SELECTOR);
+      input = wrapper.find(NUMBER_SELECTOR);
+      wrapper.setProps({ value: '12345' });
+    });
+    it('should not re-explode model', () => {
+      expect(select.props().selected.value).toEqual('+44');
+      expect(input.props().value).toBe('123456789');
+    });
+  });
+  // Prop Type will fail.
```
StevePole: If it's not possible, and there is no similar test to be written, I would delete the code.

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
+import React, { PureComponent } from 'react';
+import Types from 'prop-types';
+
+import { Select } from '../index';
+import { isValidPhoneNumber, explodeNumberModel } from './utils/pnoneNumberUtils';
+import { filterOptionsForQuery } from './utils/filterUtils';
+import countries from './data/countries';
+
+class PhoneNumberInput extends PureComponent {
+  // PropTypes definitions.
```
oliverviljamaa: I think we can get rid of these comments, as they don't add much information  🙂

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
...
+    size: 'md',
+  };
+
+  /**
+   * Initialize PhoneNumberInput
+   * @param props - object
+   */
+  constructor(props) {
+    super(props);
+    this.explodedNumber = explodeNumberModel(props.value || '', props.locale);
```
oliverviljamaa: Could we use class properties instead?

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
...
+      const { prefix, suffix } = this.explodedNumber;
+      this.setState = {
+        phoneNumberValue: suffix,
+        selectedOption: { value: prefix, label: prefix },
+        searchQuery: '',
+      };
+    }
+  }
+
+  /**
```
oliverviljamaa: I don't think these comments are necessary either 🙂

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,153 @@
...
+    this.setState({ searchQuery });
+  };
+
+  /**
+   * Retrieves filtered options based on search field value.
+   */
+  getSelectOptions = () => {
+    const options = [];
+    const filteredOptions = filterOptionsForQuery(countries, this.state.searchQuery);
+    filteredOptions.map(option =>
```
oliverviljamaa: What do you think about destructuring here?

src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -0,0 +1,189 @@
...
+describe('Given a telephone number component', () => {
+  let select;
+  let input;
+  let wrapper;
+  const props = { onChange: jest.fn() };
+  const PREFIX_SELECT_SELECTOR = 'Select';
+  const NUMBER_SELECTOR = 'input[name="phoneNumber"]';
+  describe('when initialised without a model', () => {
+    beforeEach(() => {
+      wrapper = shallow(<PhoneNumberInput {...props} />);
```
oliverviljamaa: I think the convention in this repo is `component`, or did you see `wrapper` being used?

src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -0,0 +1,189 @@
...
+      select = wrapper.find(PREFIX_SELECT_SELECTOR);
+      input = wrapper.find(NUMBER_SELECTOR);
+    });
+
+    it('should set prefix control to default UK value', () => {
+      expect(select.props().selected).toEqual({ value: '+44', label: '+44' });
+    });
+    it('should set number control to empty', () => {
```
oliverviljamaa: Some empty lines missing.

src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -0,0 +1,189 @@
...
+      expect(input.props().value).toBe('');
```
oliverviljamaa: Do you think `.prop('value')` etc would read better?

src/phoneNumberInput/utils/filterUtils.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Filters a set of options based on search string
+ * @param options
+ * @param query
+ * @returns {*}
+ */
+export const filterOptionsForQuery = (options, query) =>
```
oliverviljamaa: I like that you separated the filtering from the view files, well done!

src/phoneNumberInput/utils/filterUtils.js
```diff
@@ -0,0 +1,31 @@
...
+  !query ? options : options.filter(option => isPhoneOptionAndFitsQuery(option, query));
+
+/**
+ * Checks if query is contained into object properties.
+ * @param {object} option - the select option
+ * @param {string} query - the current search query
+ * @returns {boolean}
+ */
+export const isPhoneOptionAndFitsQuery = (option, query) =>
```
oliverviljamaa: Can you explain the rationale behind making this phone-specific?

src/phoneNumberInput/utils/pnoneNumberUtils.js
```diff
@@ -0,0 +1,94 @@
+import countries from '../data/countries';
+
+/**
+ *
+ * @param code
+ * @returns {*}
+ */
+export const findCountryByCode = code => {
+  if (code && code.length === 2) {
+    // Find longest matching prefix
```
oliverviljamaa: This comment indicates it could be extracted to a well-named function, what do you think?

src/phoneNumberInput/utils/pnoneNumberUtils.js
```diff
@@ -0,0 +1,94 @@
...
+    const matchingCodes = countries.filter(country => code === country.iso2);
+
+    if (matchingCodes && matchingCodes.length) {
+      return matchingCodes.reduce((a, b) => (a.phone.length > b.phone.length ? a : b));
+    }
+  }
+  return undefined;
```
oliverviljamaa:
```suggestion
  return null;
```

`null` indicates that it's set by us better than _defining_ something as `undefined`.

src/phoneNumberInput/PhoneNumberInput.js
AndreaTW: I think the check of the locale should be performed outside of the component. I would create a utility or a mapping object to normalize all the locale.

AndreaTW:
> Left some comments. I also noticed that its missing validations. On old version when typed asadsad for example it shows "Invalid characters"

At the moment I left the validation out of the component as @Tankenstein is currently working on all filed validations

PR #87, opened 2018-11-30
Fix Select options to bottom on XS screens

mikknarusk:
Rationale: in calculator, payment method dropdown is displayed inline, so when opened in mobile, the options will overflow out of the screen. This will add an option to display dropdown in a way that emulates apps (Android and iOS).
screen shot 2018-11-30 at 15 58 14

Feel free to suggest changes to the prop name.
Also, all dropdown are located in bootstrap, so my additions feel a bit strange in `Select.less`. I don't really think they should be in bootstrap either, so this is where they ended up.


docs/SelectDocs.js
```diff
@@ -30,6 +30,7 @@ export default class SelectDocs extends Component {
       searchValue: '',
       placeholder: 'Select an option...',
       hasClassNames: false,
+      fixedPositionOnXs: false,
```
Tankenstein: I'm not sure i'm a fan of the naming, perhaps we can brainstorm together.

Tankenstein:
In addition, i have 2 problems with it right now

a) i'm not sure the scroll behaviour is what i expect it to be, but i'll play around some more to figure this out for myself, lol
b) it's not animated when it leaves the view, this is worth fixing

docs/SelectDocs.js
mikknarusk: I agree, let's do that

oliverviljamaa: I'd be keen on avoiding the flag altogether. Why do we need the different behaviours?

mikknarusk: just discussed with @Tankenstein, I think it could be coupled to the `block` attribute, since that's the reasoning for adding this in the first place. For `block` the current behavior works fine. Unless we can agree that we could use the `fixed` option everywhere, without opt-in. Will discuss this with design as well

PR #89, opened 2018-12-08
Feature/scaffhold component

AndreaTW:


AndreaTW:
> Hey, just wondering, what are the advantages of this over an IDE snippet?

Hey, I think the advantage it's that it doesn't depend on the IDE. It can also be installed as part of dev dependencies in different project so it's easier to use. It can be published on NPM so everyone can contribute and if we update everyone will get the same update :)

taaviaasver:
Just my thoughts: I like to init components manually :innocent: and don't think it belongs to this repo

AndreaTW:
> Just my thoughts: I like to init components manually 😇 and don't think it belongs to this repo

ahaha ok, I think it would be still a cool util for those (like me :) ) who don't want to do it manually :) . Plus I still think it would be a good way to reinforce standards and consistency across all the teams :)

Which repo should I put it? :)

oliverviljamaa:
@AndreaTW I would probably use the scaffolding tool, but I don't think it belongs in this repo either, rather a separate library people could use on their command lines and globally. What do you think?

AndreaTW:
> @AndreaTW I would probably use the scaffolding tool, but I don't think it belongs in this repo either, rather a separate library people could use on their command lines and globally. What do you think?

Hey @oliverviljamaa , the idea was to publish it as a npm module and add as a dev dependency wherever we need it . Also add a new npm command to run the scaffold process 
I can move it into another repo for sure :)

PR #91, opened 2018-12-11
Feature/tw telephone component

AndreaTW:


src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+  static defaultProps = {
+    required: false,
+    disabled: false,
+    value: '',
+    locale: 'en_GB',
+    searchPlaceholder: 'Prefix',
+    size: 'md',
+  };
+
+  /* eslint-disable react/sort-comp */
```
Tankenstein: Why disable it? If you think we should disable it, then do it globally imo :)

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+  state = {
+    phoneNumberValue: (this.explodedNumber && this.explodedNumber.suffix) || '',
+    selectedOption: { value: this.explodedNumber.prefix, label: this.explodedNumber.prefix },
+    searchQuery: '',
+  };
+
+  componentDidUpdate(prevProps) {
+    if (isValidPhoneNumber(this.props.value) && prevProps.value !== this.props.value) {
+      const { prefix, suffix } = explodeNumberModel(this.props.value, this.props.locale);
+      this.setState = {
```
Tankenstein: I think think is a typo, you mean to call `this.setState({ ...stateHere })`

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+  /* eslint-disable react/sort-comp */
+  explodedNumber = explodeNumberModel(this.props.value || '', this.props.locale);
+
+  state = {
+    phoneNumberValue: (this.explodedNumber && this.explodedNumber.suffix) || '',
+    selectedOption: { value: this.explodedNumber.prefix, label: this.explodedNumber.prefix },
+    searchQuery: '',
+  };
+
+  componentDidUpdate(prevProps) {
```
Tankenstein: This should be done in `static getDerivedStateFromProps`, if i understand right what you're doing :)

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+    searchPlaceholder: 'Prefix',
+    size: 'md',
+  };
+
+  /* eslint-disable react/sort-comp */
+  explodedNumber = explodeNumberModel(this.props.value || '', this.props.locale);
+
+  state = {
+    phoneNumberValue: (this.explodedNumber && this.explodedNumber.suffix) || '',
+    selectedOption: { value: this.explodedNumber.prefix, label: this.explodedNumber.prefix },
```
Tankenstein: i would call this `selectedCountryPrefix` or something, option is not really telling me what it is in this context :)

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+    required: false,
+    disabled: false,
+    value: '',
+    locale: 'en_GB',
+    searchPlaceholder: 'Prefix',
+    size: 'md',
+  };
+
+  /* eslint-disable react/sort-comp */
+  explodedNumber = explodeNumberModel(this.props.value || '', this.props.locale);
```
Tankenstein: Do you need to keep this on the instance? I would rather you do it where you need it instead of keeping it as a property if possible

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -0,0 +1,130 @@
...
+
+          <div className="col-xs-7">
+            <div className={`input-group input-group-${size}`}>
+              <input
+                name="phoneNumber"
+                value={phoneNumberValue}
+                type="text"
+                className="form-control"
+                disabled={disabled}
+                onChange={this.onChangeInput}
```
Tankenstein: Usually, in the props we call them `onDoSomething`, but internally in a component, if you're not just passing it through, we call it like `handleSomething`, in this case probably `handleInputChange`, wdyt? Same goes for the select handlers :)

PR #94, opened 2018-12-18
Date lookup

taaviaasver:
date-lookup-demo

[Try it out](https://transferwise.github.io/components/branch/date-lookup/index.html)

Date-lookup in React, based on our [AngularJS version](https://github.com/transferwise/styleguide-components/tree/master/src/forms/date-lookup)

NB! I will update changelog and bump version after PR gets approved

Future TODO list (extras):

- handle too long date names (months in Estonian etc)
- `.today` styles (bootstrap level change)
- make it not horizontally scrollable (bootstrap level change)
- buttons vs links? (uses links, but IMO should use buttons, requires bootstrap rule change)


AndreaTW:
image

Hey i've noticed that on mobile when calendar is open is not possible to close it unless you choose a date. Is this the desired behaviour? :)

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+    return null;
+  }
+
+  element = React.createRef();
+
+  open = () => {
+    this.setState({ open: true, mode: MODE.DAY }, () => {
+      this.focusOn('.tw-date-lookup-header-current');
+    });
+    document.addEventListener('click', this.handleOutsideClick, true);
```
AndreaTW: Sorry if not relevant...but what is going to happen if the component gets SSR ? Do we need some defensive code here?

src/common/dateUtils/getDayNames/getDayNames.js
```diff
@@ -0,0 +1,11 @@
+import { formatDate } from '@transferwise/formatting';
+
+export function getDayNames(locale, format = 'short') {
```
AndreaTW: How come are you not using es6  syntax?
Is there a particular reason? 
I saw you use arrow functions somewhere else and it would be nice to keep consistency :)

src/dateLookup/DateLookup.events.spec.js
```diff
@@ -0,0 +1,135 @@
...
+      );
+    });
+  });
+
+  describe('on outside click', () => {
+    beforeEach(() => {
+      component.instance().open();
+    });
+
+    it('closes', () => {
```
AndreaTW: I don't think this is actually happening as when clicking outsde the calendar it stays open

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
+import React, { Component } from 'react';
+import Types from 'prop-types';
+
+import KeyCodes from '../common/keyCodes';
```
AndreaTW: We should add some resolvers to Webpack to avoid this folder navigation

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+  DAY: 'day',
+  MONTH: 'month',
+  YEAR: 'year',
+};
+
+const NOW = new Date();
+
+class DateLookup extends Component {
+  static propTypes = {
+    value: Types.instanceOf(Date),
```
AndreaTW: Can this be also a valid ISO string?

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+    open: false,
+    mode: MODE.DAY,
+  };
+
+  static getDerivedStateFromProps(props, state) {
+    const propsSelected = getMidnight(props.value);
+    const propsMin = getMidnight(props.min);
+    const propsMax = getMidnight(props.max);
+    const hasSelectedChanged = +state.selectedDate !== +propsSelected;
+    const hasMinChanged = +state.min !== +propsMin;
```
matallo: can't we enforce `propsMin` passed is a number in the `propTypes`? maybe I'm missing something here...

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+
+  handleOutsideClick = event => {
+    if (this.state.open && !this.element.current.contains(event.target)) {
+      this.close();
+    }
+  };
+
+  handleKeyDown = event => {
+    const { open } = this.state;
+    switch (event.keyCode) {
```
AndreaTW: This can probably be improved following the rules 

> if (event.key !== undefined) {
    // Handle the event with KeyboardEvent.key and set handled true.
  } else if (event.keyIdentifier !== undefined) {
    // Handle the event with KeyboardEvent.keyIdentifier and set handled true.
  } else if (event.keyCode !== undefined) {
    // Handle the event with KeyboardEvent.keyCode and set handled true.
  }

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#Browser_compatibility

What do you think?

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+    else if (fallbackElement) this.focusOn(fallbackElement);
+  };
+
+  switchMode = mode => {
+    this.setState({ mode }, () => {
+      this.focusOn('.active', '.today');
+    });
+  };
+
+  switchToDays = () => this.switchMode(MODE.DAY);
```
AndreaTW: I've played a bit with the component and I saw that once you click on the MM+YY link in the calendar header you end up here:
image

1) I think there's should be an easier way to get back to the days view: ie: a link at the top or similar.
2) It would be awesome to deactivate the navigation arrows if the earlier/later years are not available cause of the range

What do you think?

src/dateLookup/DateLookup.state.spec.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
AndreaTW: I love the fact that you separated this file. I'll start doing it too

src/dateLookup/DateLookup.state.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(+newState.min).toBe(+new Date(2018, 10, 27, 0, 0));
+    expect(+newState.max).toBe(+new Date(2018, 11, 28, 0, 0));
+  });
+
+  it('sets selectedDate to min when it is < min', () => {
+    const onChange = jest.fn();
+    const props = { value: new Date(2018, 11, 27), min: new Date(2018, 11, 28), onChange };
+    const newState = DateLookup.getDerivedStateFromProps(props, defaultState);
+    expect(+newState.selectedDate).toBe(+new Date(2018, 11, 28));
+    expect(onChange).toHaveBeenCalled();
```
AndreaTW: Maybe just to be picky we can verify also with values it has been called :)

src/dateLookup/DateLookup.view.spec.js
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
```
AndreaTW: Again brilliant choice to have this file separate

src/dateLookup/dayCalendar/table/DayCalendarTable.js
```diff
@@ -0,0 +1,129 @@
...
+    const weeks = this.getTableStructure();
+    return (
+      <table className="table table-condensed table-bordered table-calendar m-b-0">
+        <thead>
+          <tr>
+            {this.days.map((day, index) => (
+              <th key={day}>
+                <span className="hidden-xs">{day.substring(0, 3)}</span>
+                <span className="visible-xs-inline-block">
+                  {this.daysShort[index].substring(0, 2)}
```
AndreaTW: When changing locale the Months and the days are lowercase
image

src/dateLookup/dayCalendar/table/DayCalendarTable.spec.js
```diff
@@ -0,0 +1,125 @@
...
+      locale,
+      onSelect: jest.fn(),
+    };
+    component = shallow(<DayCalendarTable {...props} />);
+  });
+
+  it('shows weekday names in table header', () => {
+    const ths = component.find('thead th');
+    expect(ths.length).toBe(7);
+    expect(ths.at(0).text()).toContain('Mon');
```
AndreaTW: Those are going to be lowercase for different locale

src/dateLookup/tableLink/TableLink.js
```diff
@@ -0,0 +1,51 @@
...
+    event.preventDefault();
+    if (!this.props.disabled) {
+      this.props.onClick(this.props.item);
+    }
+  };
+
+  render() {
+    const { item, type, title, longTitle, active, disabled, today } = this.props;
+    return (
+      <Fragment>
```
AndreaTW: How come we have a Fragment here?

src/dateLookup/DateLookup.events.spec.js
taaviaasver: Good find, thanks.

I noticed that [v.4.9.0](https://github.com/transferwise/components/blob/master/CHANGELOG.md) introduced different dropdown visuals for small screens. It adds backdrop to `.dropdown`, which technically is inside the component (so I considered it as click inside the date-lookup).

Should be fixed now. I'm now checking if user specifically clicked on `.dropdown-menu`

Was that the case for you or did you see any other problem?

src/dateLookup/DateLookup.js
taaviaasver: Should be fine, because it is inside `open` method and will not get called on server.

`src/ssr.spec.js` also passes so I'm hoping that its ok

src/common/dateUtils/getDayNames/getDayNames.js
taaviaasver: I have no good reason for this and I'm totally ok to change..

But in my opinion its perfectly fine to use normal functions, and it will actually be few characters shorter:

`export function getDayNames(locale, format = 'short') {`

`export const getDayNames = (locale, format = 'short') => {`

So my vote goes to current solution.

For others: if you read this and have an opinion -  give thumbs up Andrea's or my comment and I'll switch to arrow functions if it has more votes.

src/dateLookup/DateLookup.js
taaviaasver: Its a bit less magic this way, so I'd prefer not to add it, but would be fine with it :)

src/dateLookup/DateLookup.js
taaviaasver: Currently - no.
But should not be that difficult to add support for it.

I just think that most of our use-cases would want to pass in `Date`-s and if anyone needs string format or something different then date-lookup can be extended.

I'm only aware of use-cases where user can check or download transactions history, for this you'd normally pass in today (`new Date()`) and 1 month ago (`let monthAgo = new Date(); monthAgo.setMonth(monthAgo.getMonth() - 1)`) as defaults

src/dateLookup/DateLookup.js
taaviaasver: Could you explain a bit more?

Currently, users can pass in `Date` object for `value`, `min` and `max` props.
Then internally I'm setting provided dates to midnight in order to make comparison easy.

`+propsMin` is used only to compare two Dates, it turns the Date to timestamp (numbers), same as `propsMin.valueOf()` would do

src/dateLookup/DateLookup.events.spec.js
AndreaTW: all good now

src/dateLookup/DateLookup.js
taaviaasver: Oh I didn't know its so messy on browsers and specs.

I used `keyCode` based on `Select` and `Popover` implementations - so same problem applies for these.

My first thought was to do something like `const key = event.key || event.keyIdentifier || event.keyCode;`, but its not going to work, because these don't return the same thing:
```
console.log(event.key, event.keyIdentifier, event.keyCode);
// -> ArrowLeft undefined 37
```

I think this issue should be addressed later in separate PR

src/dateLookup/dayCalendar/table/DayCalendarTable.js
taaviaasver: I think it should be like that, in lots of locales month names are lowercase.

I'm not constructing names myself and using browsers implementation of [toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)

src/dateLookup/dayCalendar/table/DayCalendarTable.spec.js
taaviaasver: I'm using mocked `getDayNames` (see line 14) here and can test for 'Mon' and not care about different locales.

src/dateLookup/tableLink/TableLink.js
taaviaasver: Wrapping it with `Fragment` only because I needed to add `{/* eslint-disable jsx-a11y/anchor-is-valid */}` to next line. Reason: without `Fragment` React complained about not returning a single element.

I also tried with something like this, but for some reason it didn't work:
```
return (
      <a // eslint-disable-line  jsx-a11y/anchor-is-valid 
          href=""
          onClick={this.onClick}
```

Ideally I'd go for `<button>` instead of `<a>` and remove "Fragment" and "eslint-disable" - but cannot do that yet because I'd have to update `bootstrap.css` first to style buttons correctly.

src/dateLookup/DateLookup.js
taaviaasver: I based it on our [AngularJS version](http://transferwise.github.io/styleguide-components/index.html#date-lookup), which behaves the same way and doesn't deactivate arrows.

I would not any changes right now, but these improvements can be made later

src/dateLookup/DateLookup.js
AndreaTW: The reasons why I asked is that the date field I'm working on used to accept Iso strings in the angular version. I'm not aware of all the possible use cases of the your component but if you feel confident that is not something we need now, feel free to leave it out :)

src/common/dateUtils/getDayNames/getDayNames.js
Tankenstein: Imo named functions are good practise. On their own they're even preferrable, as you can use function hoisting semantics to put them at the bottom of the file.

src/dateLookup/DateLookup.js
Tankenstein: I would rather avoid the magic as well :)

src/common/dateUtils/getDayNames/getDayNames.spec.js
```diff
@@ -0,0 +1,17 @@
+import { getDayNames } from '..';
+
+jest.mock('@transferwise/formatting', () => ({
+  formatDate: day => day.getDay(),
+}));
+
+describe('getDayNames', () => {
+  it('returns all weekdays', () => {
+    expect(getDayNames().length).toBe(7);
```
oliverviljamaa: I've used `.toBe` myself, but do you think it'd be nicer to use `.toHaveLength` instead (for slightly nicer error messages)?

src/common/dateUtils/getDayNames/getDayNames.js
Tankenstein: Named functions are good imo, especially since you can use function hoisting to move them out of your way :)

oliverviljamaa: As the module is already named after the function, I feel it'd be more intuitive to make it the default export, what do you think?

src/common/dateUtils/getDayNames/getDayNames.spec.js
Tankenstein: Cool, did not know this was a thing

src/common/dateUtils/moveDateToWithinRange/moveDateToWithinRange.js
```diff
@@ -0,0 +1,7 @@
+// Makes sure that date is between min and max dates
+export function moveDateToWithinRange(date, min, max) {
```
oliverviljamaa: I really enjoy the separation of the view and each of the utility functions.

Tankenstein: Low prio but i usually see functions like this called "clamp" something to within range, feel free to ignore though, this is also readable.

src/common/dateUtils/moveDateToWithinRange/moveDateToWithinRange.spec.js
```diff
@@ -0,0 +1,42 @@
+import { moveDateToWithinRange } from '..';
+
+describe('moveDateToWithinRange', () => {
+  it('should return null when no date is provided', () => {
```
oliverviljamaa: We normally omit the `should` these days as it's just repeated verbosity and it seems you're not doing this for every spec either. Is there a reason why we have it here?

src/dateLookup/DateLookup.events.spec.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import DateLookup from '.';
+
+describe('DateLookup (events)', () => {
```
Tankenstein: Separating these test suites is lovely.

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+    document.removeEventListener('click', this.handleOutsideClick, true);
+  };
+
+  handleOutsideClick = event => {
+    if (!this.state.open) return;
+    const dropdown = this.element.current.querySelector('.dropdown-menu');
+    if (dropdown && !dropdown.contains(event.target)) this.close();
+  };
+
+  handleKeyDown = event => {
```
oliverviljamaa: I'm wondering whether it makes sense extracting this as well for testability, WDYT?

Tankenstein:
One q: the keyboard navigation feels slow. Should we try and optimize it?

andrastothtw:
I played with it, it felt great even with keyboard! Great stuff! I'm going to try it out on my phone as well.

My two cents:
1. unnecessary selection when opened:
image

2. I can't change year with keyboard. I can click on the month-year header and then go with keyboard. The flow from there with keyboard is super good. It's just I can't select it with keyboard.

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+              />
+            )}
+          </div>
+        )}
+      </div>
+    );
+  }
+}
+
+export default DateLookup;
```
andrastothtw: Please don't use default exports: 
https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

src/common/dateUtils/getDayNames/getDayNames.js
taaviaasver: I don't have any strong opinions on that, only slight preference towards named exports, [based on this article](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad)

Started switching to defaults, then thought that its a bit clearer with named exports. So I'd like to keep it like it is..

src/dateLookup/DateLookup.js
taaviaasver: I had similar thoughts and would like to get `DateLookup.js` smaller (250 lines now).

But I didn't know whats the best way of doing it..

I thought about doing smth like that, but not sure if it works and makes sense:
```
    return (
      <div ...>
        <KeyboardEventHandler ... onKeyDown={handleKeyDown} />
        <OpenButton ... />
       ...
```

Do you have any opinions/recommendations?

src/dateLookup/DateLookup.js
taaviaasver: I'd keep it like it is for now - just because all other components are doing the same (`src/index.js`), and it can be addressed in another PR.

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+import MonthCalendar from './monthCalendar';
+import YearCalendar from './yearCalendar';
+
+const MODE = {
+  DAY: 'day',
+  MONTH: 'month',
+  YEAR: 'year',
+};
+
+const NOW = new Date();
```
mikknarusk: AFAIK, this is evaluated on page load, and only once, meaning it is global between instances of `DateLookup`. 
This can have unintended results, for example: 
- since it is used with `getMidnight `, what if day/month/year has changed from page load to user interaction
- I'm not sure how React evaluates components in SSR, but if this is anthing like regular node.js, `NOW` will be evaluated when the process (container) starts, and it will never change during its lifecycle
- relating to the last point, this will probably cause hydration problems with SSR, since `NOW` will have a different value in server vs client 

I can be totally off-base, but this has been my experience with similar stuff in Vue.js

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,249 @@
...
+    label: '',
+    shortDate: false,
+    disabled: false,
+  };
+
+  state = {
+    selectedDate: getMidnight(this.props.value),
+    min: getMidnight(this.props.min),
+    max: getMidnight(this.props.max),
+    month: getMidnight(this.props.value || NOW).getMonth(),
```
mikknarusk: Initially I thought `month` and `year` duplicate the data in `selectedDate`, but they only keep track of the UI state, right? Could we change the naming to reflect that, e.g. `visibleMonth`?

src/dateLookup/DateLookup.js
taaviaasver: Wow, nice level of details 👍 

Day/month/year change issue is valid for sure, don't know much about SSR.. I think its highly likely that there will be some mismatch on dates (SSR vs client side)

I'll try to figure something out, at least for day/month/year change issue

src/dateLookup/DateLookup.js
taaviaasver: Actually I had same thoughts when implementing it, I think I'll refactor and go for `viewMonth` and `viewYear`.

src/dateLookup/dayCalendar/table/DayCalendarTable.js
```diff
@@ -0,0 +1,129 @@
...
+
+  selectDay = day => {
+    const { year, month, onSelect } = this.props;
+    onSelect(new Date(year, month, day));
+  };
+
+  isDisabled = day => {
+    const { min, max, month, year } = this.props;
+    const date = new Date(year, month, day);
+    return !!((min && date < min) || (max && date > max));
```
mikknarusk: maybe we can extract this logic to a helper and use it in `moveDateToWithinRange ` as well

mikknarusk:
Nice! I expected to find massive uses of `moment` here, but this is awesome

src/dateLookup/DateLookup.js
```diff
@@ -0,0 +1,247 @@
...
+import MonthCalendar from './monthCalendar';
+import YearCalendar from './yearCalendar';
+
+const MODE = {
+  DAY: 'day',
+  MONTH: 'month',
+  YEAR: 'year',
+};
+
+class DateLookup extends Component {
```
AndreaTW: Also I forgot to mention...Can all of those component be and extend of PureComponent instead of normal components? 
https://codeburst.io/when-to-use-component-or-purecomponent-a60cfad01a81

taaviaasver: TBH I haven't used PureComponent-s and saw that all other components here don't use it as well.. @Tankenstein @oliverviljamaa and others - any opinions, have you considered using it?

AndreaTW: I do agree with what s written in the article. We should use PureComponents by default unless there's a specific case where we need to handle in a particular way componentShouldUpdate

taaviaasver: Thanks for the tip, I switched to PureComponent now. For my component using PureComponent makes sense. I'm a bit worried if we'd use it as default everywhere - less experienced or backend focused devs might mutate objects and run into weird issues (can this happen?! maybe this topic needs to be discussed in guild meeting or smth?)

For dateLookup biggest performance gain came from `tableLink`. It does not re-render all links anymore when moving around with arrow keys within the same year/month view.

I didn't go for PureComponent in stateless functions, `dateLookup/header/Header.js` for example, because performance gains were really tiny, but the code is more compact.
```
const Header = () => // stateless function
class Header extends PureComponent { // PureComponent
```

AndreaTW: Hey, yes maybe it's worth it to discuss it all together so we can share knowledge

Tankenstein: It's always a better idea to have a stateless function rather than a pure component, the react team has stated that afaik. Their goal is to start optimizing the stateless component case, which is also why they intro-d hooks.

PR #96, opened 2019-01-03
Initialised Date component

AndreaTW:
Initialised Date component
https://transferwise.github.io/components/branch/tw-date-component/index.html


AndreaTW:
Please find a demo here 
https://transferwise.github.io/components/branch/tw-date-component/index.html

src/date/Date.js
```diff
@@ -0,0 +1,214 @@
...
+
+import Select from '../select';
+
+import { isDateValid, getMonthNames, explodeDateModel } from './utils';
+import { DEFAULT_LOCALE } from './data';
+
+const MonthBeforeDay = ['en-US', 'ja-JP'];
+const DEFAULT_VALUES = { day: null, month: 0, year: null };
+
+class TWDate extends PureComponent {
```
Tankenstein: In react there's no need to prefix, so remove the `TW`, imo. In addition, usually we call these sorts of things `Input`, as they don't deal with formatting, but rather with input, so i would call it `DateInput`

AndreaTW: Cool I renamed cause it was clashing with the Date class :)

src/dateInput/DateInput.js
```diff
@@ -0,0 +1,214 @@
...
+      const { day, month, year, months } = this.state;
+      this.props.onChange({ day, month, year, months });
+    });
+  };
+
+  handleYearChange = event => {
+    const { value } = event.target;
+    const slicedYear = value.length > 4 ? value.slice(0, 4) : value;
+
+    if (slicedYear.length === 4) {
```
taaviaasver: Could you please explain why this block (with `checkDate`) is needed here?

My first thought is that December always has 31 days and this check is redundant

src/dateInput/DateInput.js
```diff
@@ -0,0 +1,214 @@
...
+    const { checkedDay } = this.checkDate(event.target.value, this.state.month, this.state.year);
+
+    this.setState({ day: checkedDay }, () => {
+      const { day, month, year, months } = this.state;
+      this.props.onChange({ day, month, year, months });
+    });
+  };
+
+  handleMonthChange = selectedValue => {
+    const { checkedDay } = this.checkDate(this.state.day, selectedValue.value, this.state.year);
```
taaviaasver: I noticed that this line throws error when clicking on first option (empty, "Select an option..."):
```
Uncaught TypeError: Cannot read property 'value' of null
```

src/dateInput/utils/explodeDateModel/index.js
```diff
@@ -0,0 +1,11 @@
+import { isExplodedDatePatternCorrect } from '../';
+
+export const explodeDateModel = date => {
```
taaviaasver: Just wondering.. maybe it could be `explodeDate` instead of `explodeDateModel`? I think `Model` in name does not add any useful information

src/dateInput/utils/getMonthNames/index.js
```diff
@@ -0,0 +1,14 @@
...
+export const getMonthNames = (locale, format = 'long') => {
+  const months = [];
+  const date = new Date(2018, 0, 1);
+
+  while (months.length < 12) {
+    months.push(formatDate(date, locale, { month: format }));
+    date.setMonth(date.getMonth() + 1);
+  }
+
+  return months.length ? months : MONTHS_EN;
```
taaviaasver: I think the fallback doesn't work:
`months.push(..whatever..)` always increments the length

And more on this.. maybe you could remove the fallback and add it to [`formatDate` fallback method](https://github.com/transferwise/formatting/blob/master/src/date/fallback-format/fallbackFormat.js) directly?
Currently fallback only support numeric months and ignores actual `month` format value, but it could also support `month: 'long'`. WDYT?

src/dateInput/utils/getMonthNames/index.js
```diff
@@ -0,0 +1,14 @@
+import { formatDate } from '@transferwise/formatting';
+import { MONTHS_EN } from '../../data';
+
+export const getMonthNames = (locale, format = 'long') => {
+  const months = [];
+  const date = new Date(2018, 0, 1);
+
+  while (months.length < 12) {
+    months.push(formatDate(date, locale, { month: format }));
```
taaviaasver: I recently did a performance update for `formatDate`, it would be slightly faster this way:
```
const monthFormat = { month: format };

while (months.length < 12) {
  months.push(formatDate(date, locale, monthFormat));
```
[offtopic] Unfortunately it might not be very intuitive, but I couldn't come up with any better way

src/dateInput/utils/explodeDateModel/index.js
```diff
@@ -0,0 +1,11 @@
...
+  const dateObj = typeof date === 'string' ? new Date(date) : date;
+
+  const day = dateObj.getUTCDate();
+  const month = dateObj.getUTCMonth();
+  const year = dateObj.getUTCFullYear();
+
+  return isExplodedDatePatternCorrect(year, month, day) ? { day, month, year } : null;
```
taaviaasver: Could you explain why `isExplodedDatePatternCorrect ` is needed?

Seems to be checking that every date part is positive, how can these be negative?

src/dateInput/DateInput.js
AndreaTW: Hey, I put that check to correct days in February for non-leap years. If year is not select users can enter dates like 29 February but as soon as a year is specified that value is corrected if not in a lap year

src/dateInput/utils/getMonthNames/index.js
AndreaTW: Hey , yeah I think we can probably add it into the formatDate fallback method for month:long format.

src/dateInput/utils/convertToUTCObjectDate/convertToUTCObjectDate.js
```diff
@@ -0,0 +1,15 @@
+export const convertToUTCObjectDate = date => {
+  let dateObject = date;
+
+  if (typeof date === 'string') {
+    let dateString = date.split('T')[0];
+    dateString = new Date(dateString).toISOString();
+    dateString = dateString.split('T')[0].split('-');
+    dateObject = new Date(dateString[0], dateString[1] - 1, dateString[2]);
```
taaviaasver: I think it can be simplified with:
```
let utcDate = new Date(date.split('T')[0]); // using YYYY-MM-DD creates UTC date 
dateObject = new Date(utcDate.getUTCFullYear(), utcDate.getUTCMonth(), utcDate.getUTCDate());
```

src/dateInput/utils/convertToUTCObjectDate/convertToUTCObjectDate.js
```diff
@@ -0,0 +1,15 @@
+export const convertToUTCObjectDate = date => {
```
taaviaasver: I like what you do here (ignoring times and converting to local timezone), but I think the name is a bit misleading, what do you think about renaming to smth like `convertISOStringToLocalMidnight`?

And maybe typeof string check can be excluded as well.. can we call the function only for strings?

src/dateInput/DateInput.js
```diff
@@ -0,0 +1,258 @@
...
+
+    if (isDateValid(value)) {
+      const converted = convertToUTCObjectDate(value);
+      dateObject = explodeDate(converted);
+      dateString = DateInput.getDateAsString(converted);
+    }
+
+    this.state = {
+      ...dateObject,
+      value: dateString,
```
taaviaasver: I think it would be better if you'd store JS Date Object (`const converted`) in state instead of `value: dateString`.

I think its more future proof:
- if you pass in string it would have to go through `convertToUTCObjectDate` or similar (added another comment next to this function)
- if you pass in JS Date Object, yo don't have to do anything and can just store it as value

Plus then you'd have to convert to string `DateInput.getDateAsString` only in `static returnValue`

src/dateInput/README.md
```diff
@@ -0,0 +1,20 @@
+#Usage
```
taaviaasver: IMO there should be a link in the docs like "Read more" or smth that links to this readme

src/dateInput/data/index.js
```diff
@@ -0,0 +1,16 @@
+export const MONTHS_EN = [
```
taaviaasver: Maybe it can be deleted since you're not using these as fallback anymore 🤔

PR #98, opened 2019-01-12
Refactored logic and test PhoneNumber

andreapiras:


docs/PhoneNumberInputDocs.js
```diff
@@ -3,13 +3,20 @@ import { Checkbox, PhoneNumberInput } from '../src';
...
+  { value: 'js-JP', label: 'Japanese' },
+];
 
 export default class PhoneNumberInputDocs extends PureComponent {
   state = {
     required: false,
     disabled: false,
     value: '+447573135343',
-    locale: 'en-GB',
+    locale: { value: 'en-GB', label: 'English UK' },
```
matallo: we can probably declare a constant with this, in any case I found this weird and hard to wrap my head around. I'd have approached this with a `selectedLocale` attribute, and then add a method like `getLocaleForLocaleValue`

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -34,34 +37,58 @@ class PhoneNumberInput extends PureComponent {
...
+    const cleanValue = value ? cleanNumber(value) : null;
+    const { prefix, suffix } = explodeNumberModel(cleanValue, locale);
+    const previousReturned = isValidPhoneNumber(value) ? prefix + suffix : null;
+
     this.state = {
       suffix,
       prefix,
-      value: suffix + prefix,
-      previousReturned: null,
+      previousReturned,
```
matallo: not convinced about storing suffix,  prefix, and previousReturned if the later is a transformation of the former, just store the original source, implement the transformation methods (and the tests) and happy days

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -34,34 +37,58 @@ class PhoneNumberInput extends PureComponent {
...
+      locale, // eslint-disable-line
```
matallo: why not lint this line?

src/phoneNumberInput/utils/cleanNumber/cleanNumber.js
```diff
@@ -0,0 +1,4 @@
+const DIGITS_MATCH = /^$|^(\+)|([\d]+)/g;
```
matallo: niiice having a separate utils

src/phoneNumberInput/PhoneNumberInput.js
andreapiras: When the state is used only inside static methods for some reason the linter complains about 'variable assigned but never used". After refactoring though it won't be necessary

src/phoneNumberInput/PhoneNumberInput.js
andreapiras: I've removed it, not sure if the implementation got easier. But I agree that the less state variables the better it is

PR #102, opened 2019-01-25
Add an XL version of the modal component

codewhisperer:
I need this to display the new rate graph on the home page and I though this might be useful here.

image


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v4.14.0
+## Allow Select to open upwards
```
KarlEST: copypasta

PR #104, opened 2019-01-28
feat: implement InstructionsList component

pterblgh:
image
localhost_3000_ iphone 6_7_8 2


StevePole:
One comment on interface, this encodes the assumption that 'donts' always follow 'dos'.  I _think_ that's probably correct, but is it something you verified with design?

pterblgh:
Validated with design, also renamed the component to `InstructionList`.

PR #106, opened 2019-02-01
Fix flaky getMidnight test

taaviaasver:
`getMidnight` tests failed for users in GMT +X timezones.


src/common/dateUtils/getMidnight/getMidnight.spec.js
```diff
@@ -11,9 +11,14 @@ describe('getMidnight', () => {
     expect(getDatePartsArray(date)).toEqual([1995, 11, 17, 0, 0]);
   });
 
-  it('works for date in format "YYYY-MM-DDTHH:SS:SS"', () => {
-    const date = getMidnight(new Date('1995-01-01'));
-    expect(getDatePartsArray(date)).toEqual([1995, 0, 1, 0, 0]);
+  it('works for date in format "YYYY-MM-DD" (UTC)', () => {
+    const input = new Date('1995-01-01');
+    const output = getMidnight(input);
+    expect(output.getFullYear()).toBe(input.getFullYear());
```
StevePole: Better to test 1995, as your test is dependent on the implementation of this function working as you expect

src/common/dateUtils/getMidnight/getMidnight.spec.js
```diff
@@ -11,9 +11,14 @@ describe('getMidnight', () => {
     expect(getDatePartsArray(date)).toEqual([1995, 11, 17, 0, 0]);
   });
 
-  it('works for date in format "YYYY-MM-DDTHH:SS:SS"', () => {
-    const date = getMidnight(new Date('1995-01-01'));
-    expect(getDatePartsArray(date)).toEqual([1995, 0, 1, 0, 0]);
+  it('works for date in format "YYYY-MM-DD" (UTC)', () => {
```
StevePole: Why UTC?  This test is not using UTC function, new Date will not give a UTC date (unless you're in UTC), and getHours (or any of the rest) will not return UTC (also unless you happen to be in UTC).  The tests work because both sets of functions are using localised times.

taaviaasver: JS treats dates created in "YYYY-MM-DD" format as UTC, doesn't matter on the timezone.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#ECMAScript_5_ISO-8601_format_support

And this is the reason why it fails and I can't just check `toBe(1995)`.
On GMT +{whatever} output is going to be 1994, on GMT -{whatever} its 1995.

StevePole: But in that case I don't think you're testing anything useful.  It would be better to test
`expect(output.getUTCFullYear()).toBe(1995);` as then you're actually testing the value of output, not just that input & output happen to return the same result for a function who's behaviour varies by timezone.

StevePole: Also, just tested in a couple browsers
Chrome:
image
Safari:
image

StevePole: Ah, actually I guess that it UTC 0...

StevePole: Yep
image

taaviaasver: I'd like to test against fixed value but can't: `output.getUTCFullYear()` on GMT -X is 1994, not 1995
```
// Los Angeles, GMT-0800 (Pacific Standard Time)
const input = new Date('1995-01-01'); // -> 31 Dec 1994 16:00
const output = getMidnight(input); // -> 31 Dec 1994 00:00
expect(output.getFullYear()).toBe(1995);  // -> fails, expects 1994
expect(output.getUTCFullYear()).toBe(1995);  // -> fails, expects 1994
```
**NB!:** I actually did a mistake in earlier comments, output will be 1994 in **GMT -X** timezones, not **GMT +X**.

I'm now actually thinking about just deleting this test, as its confusing and I don't see people using it that way, because:

1. `YYYY-MM-DD` will have UTC time 00:00 anyway
2. there could/should be similar `getUTCMidnight` function when anyone wants to use UTC

What do you think?

PR #108, opened 2019-02-14
FormControl component

pterblgh:
## Title
FormControl component

## Description
This component is a form control that can be fully configured based on a set of parameters. The control can be used as the basis for a fully dynamic form based on an external specification.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## What's missing?

- Validation (required, minLength, maxLength, min, max, pattern) - *it should be handled in Field*
- ~Upload component~
- ~`textFormat` functionality~
- async validation functionality ( we decide to implement it in a second iteration )


tiborsaas-tw:
Just by looking at the docs code I'm confused a bit how this should render. Should this component display a group of form elements? What use case does it solve?

pterblgh:
FormControl is a low level UI component which will be used by Field component which is basically responsible to map a Dynamic Forms JSON definition to a form element. 

FormControl is just a helper component to make Field more lightweight in the future by decoupling  exact input type rendering.

This component is the first step of Dynamic Flows re-write in React.

docs/FormControlDocs.js
```diff
@@ -0,0 +1,227 @@
...
+              </div>
+
+              {this.state.type === FormControlType.NUMBER && (
+                <Fragment>
+                  <div className="m-t-2">
+                    <label htmlFor="numberMin">Min</label>
+                    <input
+                      name="numberMin"
+                      className="form-control"
+                      type="number"
```
andreapiras: this doesn't work on FF :)

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,370 @@
+import React, { Component, Fragment } from 'react';
+import Types from 'prop-types';
+import { FormControlType } from './FormControlType';
+import Radio from '../radio';
+import Select from '../select';
+import DateInput from '../dateInput/DateInput';
+import PhoneNumberInput from '../phoneNumberInput/PhoneNumberInput';
+import Checkbox from '../checkbox/Checkbox';
+
+export default class FormControl extends Component {
```
andreapiras: Can we use PureComponent here?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,370 @@
...
+    minLength: null,
+    maxLength: null,
+    defaultValue: undefined,
+    searchPlaceholder: null,
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.handleOnFocus = this.handleOnFocus.bind(this);
```
andreapiras: Is there any reason why we are using those instead of arrow functions?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,370 @@
...
+    this.handleOnBlur = this.handleOnBlur.bind(this);
+  }
+
+  state = {
+    selectedOption: undefined,
+    value: this.props.defaultValue || undefined,
+    touched: false,
+  };
+
+  static getUniqueId() {
```
andreapiras: Can this be a util function?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,370 @@
...
+    FormControl.UNIQUE_ID += 1;
+    return `${FormControl.UNIQUE_ID}`;
+  }
+
+  /**
+   * autocomplete hides our form help so we need to disable it when help text
+   * is present. Chrome ignores autocomplete=off, the only way to disable it is
+   * to provide an 'invalid' value, for which 'disabled' serves.
+   */
+  getAutocompleteStatus() {
```
andreapiras: This is just stylistic but it can be rewritten as getAutocompleteStatus = () => this.props.helpText ? 'disabled' : 'on' which is more concise

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,370 @@
...
+    if (option.secondary) {
+      classes = `${classes} radio-lg`;
+    }
+    return classes;
+  }
+
+  handleOnChange(event) {
+    const { onChange } = this.props;
+    if (!event || !event.target) {
+      if (event.label && event.value) {
```
andreapiras: This logic is a bit weird...if !event ---> get into the if---> check event.label and event.value

andreapiras: Do you think it can be rewritten to something like that? ```const handleOnChange = (event) => {
  const { onChange } = this.props;
  let value = '';
  if (event) {
    if (!event.target) {
      value = event;
      if (event.value) {
        ({value} = event)
      }
    } else {
      ({value} = event.target)
    }
    return onChange(value);
  }
}```

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,362 @@
...
+      }
+      this.setState({ value: event });
+      return onChange(event);
+    }
+
+    this.setState({ value: event.target.value });
+    return onChange(event.target.value);
+  };
+
+  handleOnFocus = event => {
```
andreapiras: handleOnFocus = event => this.props.onFocus(event);

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,362 @@
...
+
+  renderField() {
+    const {
+      name,
+      id,
+      label,
+      placeholder,
+      step,
+      locale,
+      type,
```
andreapiras: I saw we have type number which is not supported in FF. I would not rely on browser implementation yet

pterblgh: According to CanIUse and to a CodePen test,  FF do support `<input type="number" />`.
https://caniuse.com/#search=type%3D%22number%22

andreapiras: Yeah I checked that too, but I opened it with my FF (64.0.2) and I was able to type letters

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+    } = this.props;
+    const { value } = this.state;
+    switch (type) {
+      case FormControlType.RADIO: {
+        return (
+          <Fragment>
+            {options.map((option, index) => (
+              // eslint-disable-next-line react/no-array-index-key
+              <Fragment key={index}>
+                <Radio
```
andreapiras: Do you think we can separate those into their own files and do do something like ---> return getRadio(this)

andreapiras: So we can keep the file lean and probably it would be easier even to split tests

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+            autoComplete={this.getAutocompleteStatus()}
+          />
+        );
+      }
+      case FormControlType.DATE:
+      case FormControlType.DATETIME: {
+        return (
+          <DateInput
+            disabled={disabled}
+            size="md"
```
andreapiras: should size be a prop?

src/FormControl/FormControlType.js
```diff
@@ -0,0 +1,29 @@
+export const RADIO = 'radio';
```
andreapiras: Do we need to export them individually?

src/FormControl/index.js
```diff
@@ -0,0 +1,3 @@
+import FormControl from './FormControl';
```
andreapiras: this can be replace with export { default as FormControl } from './FormControl';

src/select/Select.js
```diff
@@ -98,6 +104,12 @@ export default class Select extends Component {
     }, 0);
   }
 
+  handleOnFocus = event => this.props.onFocus(event);
+
+  handleOnBlur = event => {
```
andreapiras: handleOnBlur = event =>  this.props.onBlur(event);

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+    options: [],
+    step: 1,
+    // uploadAccept: false,
+    // uploadIcon: null,
+    // uploadTooLargeMessage: null,
+    // helpImage: null,
+    disabled: false,
+    readOnly: false,
+    required: true,
+    onBlur() {},
```
andreapiras: Nice never saw this

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+  };
+
+  handleOnChange = event => {
+    if (!event) {
+      return this.setStateAndOnChange(event);
+    }
+
+    if (!event.target) {
+      // This is coming from either <Select /> or <Radio />
+      if (event.label && event.value) {
```
andreapiras: Why do u have to check on the event.label if you don't use it?

mikknarusk:
Played around with it, nice work! 

One Q: do we want the textarea to be resizable horizontally?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+import PhoneNumberInput from '../phoneNumberInput/PhoneNumberInput';
+import Checkbox from '../checkbox/Checkbox';
+
+export default class FormControl extends PureComponent {
+  static DISABLED_LABEL_TYPES = [FormControlType.CHECKBOX];
+  static UNIQUE_ID = 0;
+
+  static propTypes = {
+    type: Types.oneOf([
+      FormControlType.RADIO,
```
mikknarusk: I'd maybe export something like `FormControlTypes` or `FormControlTypeValues` that's generated from `FormControlType` using `Object.keys`. Unless there's an argument for controlling these values one-by-one

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+    max: null,
+    minLength: null,
+    maxLength: null,
+    defaultValue: undefined,
+    searchPlaceholder: null,
+  };
+
+  state = {
+    selectedOption: undefined,
+    value: this.props.defaultValue || undefined,
```
mikknarusk: why the `|| undefined`?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+    touched: false,
+  };
+
+  static getUniqueId() {
+    FormControl.UNIQUE_ID += 1;
```
mikknarusk: I'm thinking how this will behave with SSR... Can't confirm any of this but will the order of rendering components be deterministic in SSR vs client-side rendering?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,367 @@
...
+  /**
+   * autocomplete hides our form help so we need to disable it when help text
+   * is present. Chrome ignores autocomplete=off, the only way to disable it is
+   * to provide an 'invalid' value, for which 'disabled' serves.
+   */
+  getAutocompleteStatus = () => (this.props.helpText ? 'disabled' : 'on');
+
+  getRadioClassNames(option) {
+    let classes = '';
+    if (this.props.disabled) {
```
mikknarusk: this repo already includes `classnames`, I'd use that here

mikknarusk: actually this method seems unused?

pterblgh: Yeah, it was unused, just removed it.

src/FormControl/FormControl.js
pterblgh: I wouldn't go with this approach, instead maybe create a `<RadioGroup />` component which accepts options and all the necessary inputs and outputs as callbacks?

src/FormControl/FormControl.js
pterblgh: As validation is not in the scope of this current PR I think it's okay to go with this approach for now and re-visit this topic when we're implementing proper validation in Field component. What do you think?

andreapiras: Hey yeah, I do agree this should be addressed as part of validation component

src/FormControl/FormControl.js
pterblgh: Updated the component in https://github.com/transferwise/components/pull/108/commits/f8918ad2e8fa5c8172e03ab0f2a3f8446256a33c.

Also, refactored the logic to the `common` folder and reused that in Tooltip too, see https://github.com/transferwise/components/pull/108/commits/d5f4e8538eb58ccd6e6f22827ccaee7f49698b3a

src/FormControl/FormControl.js
pterblgh: Went with another approach due to SSR reasons, if `id` is not passed in as a prop, we generate a pseudo-random unique for it on the client side.

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,354 @@
...
+export default class FormControl extends PureComponent {
+  static DISABLED_LABEL_TYPES = [FormControlType.CHECKBOX];
+
+  static propTypes = {
+    type: Types.oneOf(Object.values(FormControlType)),
+    name: Types.string.isRequired,
+    id: Types.string,
+    label: Types.string,
+    placeholder: Types.string,
+    helpText: Types.string,
```
StevePole: Why is helpText part of the interface, shouldn't that be at field level?

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,354 @@
...
+    // helpImage: Types.string,
+    options: Types.arrayOf(
+      Types.shape({
+        label: Types.string.isRequired,
+        value: Types.oneOfType([Types.number, Types.string, Types.bool, Types.object]).isRequired,
+        secondary: Types.string,
+      }),
+    ),
+    disabled: Types.bool,
+    readOnly: Types.bool,
```
StevePole: disabled and required match the naming of the HTML standard.  The equivalent for read only is 'readonly' not readOnly, so this felt a little jarring to me.

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,354 @@
...
+    required: Types.bool,
+    onChange: Types.func.isRequired,
+    onBlur: Types.func,
+    onFocus: Types.func,
+    minLength: Types.number,
+    maxLength: Types.number,
+    min: Types.number,
+    max: Types.number,
+    // pattern: Types.instanceOf(RegExp),
+    defaultValue: Types.oneOfType([Types.string, Types.number, Types.bool, Types.object]),
```
StevePole: Not sure form control should have defaultValue.  I think it's fine for it to have a single 'value' property.  The consumer may choose to pass in an existing value, or a default if it decides that's appropriate.

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,354 @@
...
+    if (event) {
+      if (!event.target) {
+        value = event;
+        if (event.value) {
+          ({ value } = event);
+        }
+      } else {
+        ({ value } = event.target);
+      }
+    }
```
StevePole: Found the above block hard to read.

src/FormControl/FormControl.js
```diff
@@ -0,0 +1,354 @@
...
+        );
+      }
+    }
+  }
+
+  render() {
+    const { label, id } = this.props;
+    return label && this.showLabel() ? (
+      <Fragment>
+        <label htmlFor={id}>{label}</label>
```
StevePole: Wasn't expecting to see label within form-control.  Why isn't that part of field?

src/checkbox/Checkbox.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import Types from 'prop-types';
 import CheckboxButton from '../common/CheckboxButton';
 
-const Checkbox = ({ checked, required, disabled, label, onChange }) => {
+const Checkbox = ({ checked, required, disabled, label, onFocus, onChange, onBlur }) => {
```
StevePole: Order of these new events seems weird?  Not alphabetic, onFocus and onBlur are closely related and optional

src/common/CheckboxButton/CheckboxButton.js
```diff
@@ -2,14 +2,14 @@ import React from 'react';
 import Types from 'prop-types';
 import classNames from 'classnames';
 
-const CheckboxButton = ({ checked, onClick, disabled, className }) => (
+const CheckboxButton = ({ checked, onFocus, onClick, onBlur, disabled, className }) => (
```
StevePole: Ditto, ordering.  Let's stay consistent with the above.

src/common/RadioButton/RadioButton.js
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 
 import './RadioButton.less';
 
-const RadioButton = ({ id, name, checked, onChange, disabled }) => (
+const RadioButton = ({ id, name, checked, onFocus, onChange, onBlur, disabled }) => (
```
StevePole: Ordering

src/FormControl/FormControl.spec.js
```diff
@@ -0,0 +1,664 @@
...
+    it('should render two radio buttons', () => {
+      expect(radios).toHaveLength(2);
+    });
+
+    it('should use the options correctly for the label', () => {
+      expect(radios.get(0).props.label).toBe('One');
+      expect(radios.get(1).props.label).toBe('Two');
+    });
+
+    xit('should use secondary label correctly', () => {
```
StevePole: Would like it if we could implement this if it's not.  I've seen a requirement for it.

src/FormControl/FormControl.spec.js
```diff
@@ -0,0 +1,664 @@
...
+    afterEach(() => {
+      component.unmount();
+    });
+
+    it('should render a checkbox input', () => {
+      expect(checkbox.exists()).toBeTruthy();
+    });
+
+    testFocusHandler(() => {
+      checkbox.find('button').simulate('focus');
```
StevePole: (This is definitely my fault!)
What should really happen here, is that the checkbox should be a mock component, we should test what happens when the mock component fires it's onFocus handler, and make sure that the FormControl triggers it's onFocus handler in turn.  Right now, if the implementation of the checkbox component were to change, and it no longer contained a button, it would cause tests failures in this (unrelated) component.

The same thing applies in quite a lot of places in this file for Change and blur events across different child components.

src/FormControl/FormControl.spec.js
```diff
@@ -0,0 +1,664 @@
...
+    });
+
+    xit('should use secondary label correctly', () => {
+      expect(radios[1].classList).toContain('radio-lg');
+      const smallLabel = radios[1].find('small');
+      expect(smallLabel.text()).toBe('Secondary label');
+    });
+
+    testFocusHandler(() => {
+      radio.find('input[type="radio"]').simulate('focus');
```
StevePole: Similar thing regarding mock components and callbacks.

src/FormControl/FormControl.spec.js
```diff
@@ -0,0 +1,664 @@
...
+      });
+
+      if (placeholder) {
+        it(`should render the ${controlType} control with a placeholder`, () => {
+          expect(input.prop('placeholder')).toBe(placeholder);
+        });
+      }
+
+      testFocusHandler(() => {
+        input.simulate('focus');
```
StevePole: The same mock component/event callback issue.

src/FormControl/FormControl.spec.js
```diff
@@ -0,0 +1,664 @@
...
+        input = component.find(selector);
+      });
+
+      it(`should render a ${controlType} control`, () => {
+        expect(input.exists()).toBeTruthy();
+      });
+
+      if (placeholder) {
+        it(`should render the ${controlType} control with a placeholder`, () => {
+          expect(input.prop('placeholder')).toBe(placeholder);
```
StevePole: This is an issue related to the mock component/callback thing.  We're testing the implementation for the component, we should really only test that the placeholder has been passed to the child component (that we mocked).

src/FormControl/FormControl.js
pterblgh: I've choosen the React naming convention for this prop, I think it makes sense to match it with our own higher level components.

https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes

src/FormControl/FormControl.js
pterblgh: `helpText` is [present on the field object](https://github.com/transferwise/styleguide-components/blob/164358e6a367fb6e103612217e1f0ff3c7cc7671/src/forms/field/field.html#L16) and is present in current [twFormControl](https://github.com/transferwise/styleguide-components/blob/14185513a6266b3cb53f0bd806684c09da233f5e/src/forms/form-control/form-control.component.js#L16) implementation as well and used for determining wether we should enable autocomplete for an `<input />` element or not.

I think it matches the same purpose and functionality with `getAutocompleteStatus` method:

https://github.com/transferwise/components/blob/d161487f3e56fc9a8b398dd036126ed358bddaa9/src/FormControl/FormControl.js#L99-L104

src/FormControl/FormControl.js
pterblgh: ESLint gives an error for this, I am okay with disabling the rule for these lines.

```js
// Use object destructuring.eslint(prefer-destructuring)
value = event.value;
```
On the other hand, there are some other components where this same pattern is used:

https://github.com/transferwise/components/blob/ebce9f41e2c8502ba176a75d09fc21337864e345/src/phoneNumberInput/PhoneNumberInput.js#L99

Anyway, I am happy to change it, but I think we should avoid having comments in the code  disabling ESLint rules. What do you think?

src/FormControl/FormControl.js
pterblgh: Thanks for pointing this out, I'll remove it, it's the responsibility of `<Field />`

src/checkbox/Checkbox.js
pterblgh: How about this?

- First group of parameters are "static" props which just drives the component's "state" (in the sense of UI representation)
- Second group of parameters are "dynamic" or callback related props - `onChange, onFocus, onBlur` would be a good order?

src/FormControl/FormControl.js
StevePole: Ok, I think we can reduce coupling by having a prop specifically for autocomplete.

src/FormControl/FormControl.js
StevePole: It's not so much the use of syntax (destructuring), more that the block from line 118 to 127 is just tough to read, and hard to understand what it's achieving.  Maybe you can make it more readable.

src/checkbox/Checkbox.js
StevePole: Sounds good 👍

docs/FormControlDocs.js
```diff
@@ -0,0 +1,222 @@
...
+      value,
+      searchPlaceholder,
+    } = this.state;
+    return (
+      <div className="container">
+        <section className="section">
+          <div className="row m-t-4">
+            <div className="col-md-6">
+              <h2>FormControl</h2>
+              <p>To render atomic form elements driven by Dynamic Forms API schema</p>
```
Tankenstein: This is the absolute dumbest comment ever and feel free to ignore, but we have always just put some dumb pun as the tagline, for example, "America's next top modal" under the modal component 😄

src/FormControl/FormControl.js
pterblgh: I've introduced `autoComplete` boolean prop which should drive this functionality instead of the presence of `helpText`.

src/FormControl/FormControl.js
pterblgh: I've tried to refactor it, please have a look on that method.

src/FormControl/FormControl.spec.js
pterblgh: Added the implementation to `<Radio />`, covered with test cases in that component as well, in `<FormControl />` I'm just testing wether the prop has been passed down to the `<Radio />` component.

https://github.com/transferwise/components/blob/023e4f5b0ceaa45a2780a604f1b0243b6b412fe2/src/radio/Radio.js#L7-L16

StevePole:
This is now looking good apart from the test stuff I mentioned.  I'm willing to retract my "request changes" if you'd like to push it through, as the coverage is there.  Although I do think it would be nice to isolate the tests through more extensive mocking.

src/formControl/FormControl.docs.js
```diff
@@ -0,0 +1,192 @@
...
+                uploadProps={{ usDisabled: disabled, size: SizeType.MD }}
+              />
+              <div className="m-t-2">
+                {this.state.value && <pre>{JSON.stringify(this.state.value)}</pre>}
+              </div>
+              <div className="m-t-4">
+                {[FormControlType.TEXT, FormControlType.TEXTAREA, FormControlType.NUMBER].includes(
+                  type,
+                ) && (
+                  <div className="row">{KNOBS.textKnob.map(knob => generateInput(knob, this))}</div>
```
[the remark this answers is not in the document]
dozortsev: and I think that we don't need them here, this feature it's sort of already out of scope of FormControl component.., in this docs I would just showcase the features of the abstraction (FormControl) e.g the ability to render certain component based on provided type, does this make sense?

src/formControl/FormControl.js
```diff
@@ -0,0 +1,402 @@
...
+    searchPlaceholder: null,
+    size: SizeType.MD,
+    uploadProps: {},
+    textFormatPattern: null,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      id: props.id || uniqueId(),
```
dozortsev: why we need `uniqueId()` here? and I think that `id` shouldn't be in _state_

src/common/SizeType.js
```diff
@@ -0,0 +1,6 @@
+export const SizeType = {
+  XS: 'xs',
+  SM: 'sm',
+  MD: 'md',
+  LG: 'lg',
+};
```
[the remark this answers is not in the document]
dozortsev: we can start using this in other components like _Alert_, _Button_...instead of _AlertSize_ / _ButtonSize_..

docs/utils/generateInput/generatInput.js
```diff
@@ -37,6 +38,21 @@ const generateInput = (knobs, componentThis) => {
...
+        <div className="col-md-6 m-b-2" key={state}>
+          <label htmlFor={state} className="control-label">
+            {label}
+          </label>
+          <input
+            id={state}
+            type="number"
+            className="form-control"
+            value={componentThis.state[state]}
+            onChange={ev => componentThis.setState({ [state]: parseFloat(ev.target.value) })}
```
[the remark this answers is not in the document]
andreapiras: Sorry I don't get this comment :) Why do you want to use InputWithTextFormat here?

src/common/SizeType.js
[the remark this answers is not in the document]
andreapiras: What about ```SIZES```

src/formControl/FormControl.docs.js
```diff
@@ -0,0 +1,192 @@
...
+  };
+  handleOnBlur = () => {
+    console.log('[FormControl] onBlur called');
+  };
+  handleOnFocus = () => {
+    console.log('[FormControl] onFocus called');
+  };
+
+  handleSelectChange = selValue => {
+    let newValue = '';
```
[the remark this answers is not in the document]
andreapiras: null is not suitable for input text it would throw a console warning :)

src/formControl/FormControl.docs.js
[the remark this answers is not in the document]
andreapiras: Yep I agree

src/formControl/FormControl.js
```diff
@@ -0,0 +1,402 @@
...
+          disabled,
+          autoComplete: this.getAutocompleteStatus(),
+        };
+
+        if (this.props.textFormatPattern) {
+          return (
+            <InputWithTextFormat textFormatPattern={this.props.textFormatPattern} {...inputProps} />
+          );
+        }
+        return <input {...inputProps} />;
```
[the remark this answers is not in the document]
andreapiras: InputWithTextFormat has a lot of other functionalities attached that are not needed so I rather keep the use of plain input/textarea

src/formControl/FormControl.js
andreapiras: tbh I don't know @pterblgh Is there any reason behind this?

docs/utils/generateCodeBlock/generateCodeBlock.js
```diff
@@ -1,35 +1,35 @@
 import React from 'react';
 
 const generateCodeBlock = (componentName, knobs, componentThis, extraPropsDocs = []) => {
-  const docs = Object.values(knobs).map(knob =>
-    generateDocsWithUpdatedStateValue(knob, componentThis).join('\n\r  '),
-  );
+  const arrayDocs = [];
+  Object.values(knobs).forEach(knob => arrayDocs.push(...knob));
+
+  const docs = arrayDocs.map(knob => generateDocsWithUpdatedStateValue(knob, componentThis));
```
rs6g10: Why are you iterating over arrayDocs twice? forEach and map? You can do it in one pass, no?
Also what is this naming `componentThis`? Is it the lexical scope variable?

src/formControl/FormControl.docs.js
```diff
@@ -0,0 +1,192 @@
...
+  { value: FormControlType.UPLOAD, label: 'Upload' },
+];
+
+const KNOBS = {
+  sharedKnob: [
+    { type: 'checkbox', label: 'Is Control disabled?', state: 'disabled', defaultState: false },
+    { type: 'checkbox', label: 'Is Control readOnly?', state: 'readOnly', defaultState: false },
+    { type: 'checkbox', label: 'Is Control required?', state: 'required', defaultState: false },
+  ],
+  numberKnob: [
```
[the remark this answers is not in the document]
andreapiras: As we discussed I removed them

docs/utils/generateInput/generatInput.js
rs6g10: Usually I will take this out as a method and assign that method to onChange. This way I can read and test it better.

src/common/unique-id.js
```diff
@@ -0,0 +1,4 @@
+export const uniqueId = () =>
+  Math.random()
+    .toString(36)
+    .substring(7);
```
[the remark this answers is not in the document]
rs6g10: I agree with Anton. Use a uuid library.

docs/utils/generateInput/generatInput.js
```diff
@@ -3,6 +3,17 @@ import { Select, Checkbox } from '../../../src';
...
   const { type, label, state, options } = knobs;
+
+  const handleOnChangeSelect = selection =>
+    componentThis.setState({ [state]: selection ? { ...selection } : undefined });
+
+  const handleOnChange = ev => componentThis.setState({ [state]: ev.target.value });
+
+  const handleOnChangeCheckbox = value => componentThis.setState({ [state]: value });
+
+  const handleOnChangeNumber = ev =>
```
dozortsev: let's call Event either `e` or `event`, ok? :)

docs/utils/generateInput/generatInput.js
```diff
@@ -32,12 +40,28 @@ const generateInput = (knobs, componentThis) => {
...
+            {label}
+          </label>
+          <input
+            id={state}
+            type="number"
+            className="form-control"
+            value={componentThis.state[state]}
+            onChange={handleOnChangeNumber}
+          />
+        </div>
```
dozortsev: I think we should use our components in our docs, e.g in this case we should use `<InputWithTextFormat /> ` instead native input HTML tag because our component has already implemented our CSS approach (only we only use `form-control` class but what if we decided to change it then will need to update such places..) and e.g has implemented accessibility things (even if it's not, now, but it surely going to happen in the future).

dozortsev: I think label tag should wrap element which it belong to

src/common/Sizes.js
```diff
@@ -0,0 +1,6 @@
+export const SIZES = {
+  XS: 'xs',
+  SM: 'sm',
+  MD: 'md',
+  LG: 'lg',
```
dozortsev: don't you think that names like `X_SMALL` , `LARGE` ... are more readable and better in general rather this acronyms?

src/formControl/FormControl.docs.js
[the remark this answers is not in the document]
dozortsev: then why we have this code here?

src/formControl/FormControl.docs.js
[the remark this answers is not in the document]
dozortsev: ok

src/formControl/FormControl.js
```diff
@@ -0,0 +1,405 @@
...
+    onChange: Types.func.isRequired,
+    onBlur: Types.func,
+    onFocus: Types.func,
+    minLength: Types.number,
+    maxLength: Types.number,
+    min: Types.number,
+    max: Types.number,
+    textFormatPattern: Types.string,
+    value: Types.oneOfType([Types.string, Types.number, Types.bool, Types.object]),
+    // validationAsync: Types.func
```
dozortsev: could you leave comment that we going to implement this feature a bit later or leave some sort of TODO?

src/formControl/FormControl.js
dozortsev: but now it's actually your PR and you have to figure out :)

as I mention above I don't see the use-case where id should be as part of React compo nent's state, thus we can try to remove it out from there

src/formControl/FormControl.js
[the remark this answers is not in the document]
dozortsev: > ...has a lot of other functionalities attached that are not needed...

for example?

src/formControl/FormControl.js
```diff
@@ -0,0 +1,405 @@
...
+            required={required}
+            disabled={disabled}
+            placeholder={placeholder}
+            searchPlaceholder={searchPlaceholder}
+          />
+        );
+      }
+      case FormControlType.NUMBER: {
+        return (
+          <input
```
dozortsev: same question :) why not `<InputWithTextFormat />`? we might would need test-formatting feature for number input as well, no? wdyt?

src/textareaWithTextFormat/TextareaWithTextFormat.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Types from 'prop-types';
+import WithTextFormat from '../withTextFormat';
+
+const TextareaWithFormat = props => (
```
dozortsev: don't see tests for this component

docs/utils/generateInput/generatInput.js
andreapiras: > I think we should use our components in our docs, e.g in this case we should use `<InputWithTextFormat /> ` instead native input HTML tag because our component has already implemented our CSS approach (only we only use `form-control` class but what if we decided to change it then will need to update such places..) and e.g has implemented accessibility things (even if it's not, now, but it surely going to happen in the future).

Yeah but in this case those are generating knobs for the documentation and is not required any TextFormat functionality so I rather keep the native input :)

src/formControl/FormControl.js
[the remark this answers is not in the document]
andreapiras: TextFormat render props attaches method to the rendered element which listen on onChange event and update the value of the input according to a pattern plus recalculate the cursor position on every keydown. For as easy as it seems it's quite lot of overhead compare to a native input so I don't see why introducing functionalities that are not used, I rather use native input that also we r sure that it will be less likely to introduce bugs :)

src/common/Sizes.js
andreapiras: I personally prefer those as they are in sync we what's currently used in bootstrap and people are quite familiar with

src/formControl/FormControl.js
andreapiras: You can't introduce anything but numbers in an input number so if you need something like a sort code input for example u should use a input type text

src/formControl/FormControl.docs.js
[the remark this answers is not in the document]
andreapiras: Because those are still used to generate the code block on the left :)

src/formControl/FormControl.js
pterblgh: The `id` prop can be used to match with an appropriate `<label>` element living outside of `FormControl`. Also, it is historically coming from the styleguide-components but I don't see any other valid usage over there as well. 🤔

docs/utils/generateInput/generatInput.js
```diff
@@ -3,6 +3,17 @@ import { Select, Checkbox } from '../../../src';
 
 const generateInput = (knobs, componentThis) => {
   const { type, label, state, options } = knobs;
+
+  const handleOnChangeSelect = selection =>
+    componentThis.setState({ [state]: selection ? { ...selection } : undefined });
+
+  const handleOnChange = event => componentThis.setState({ [state]: event.target.value });
+
+  const handleOnChangeCheckbox = value => componentThis.setState({ [state]: value });
```
dozortsev: what about `handleOnClick` name?

src/common/Sizes.js
dozortsev: yes, values are in sync with BT make sense keep acronyms for them but name of props should be whole words I think, because we have similar "component config" constants and there we use whole word, e.g [Button's `Danger`, `Secondary`..](https://github.com/transferwise/components/blob/2e551abc13cda2b602d5e6cc1d2069db943d5862/src/button/Button.js#L6-L24), [Alert's `UpLeft`..., `Success`](https://github.com/transferwise/components/blob/82febb14b3dc86f4530f3a103e9b9b1d7b064ad9/src/alert/Alert.js#L6-L25) etc...

src/common/unique-id.js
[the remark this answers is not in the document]
dozortsev: so can we use some lib here?

src/formControl/FormControl.docs.js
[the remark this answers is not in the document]
dozortsev: I'm confused but anyway... :)

src/formControl/FormControl.js
dozortsev: do you mean that `<InputWithTextFormat />` doesn't support only numbers?

src/formControl/FormControl.js
```diff
@@ -0,0 +1,403 @@
...
+            readOnly={readOnly}
+            required={required}
+            step={step}
+            type="number"
+            value={value}
+          />
+        );
+
+      case FormControlType.HIDDEN:
+        return <input type="hidden" name={name} value={value} />;
```
dozortsev: `id`?

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,14 @@
...
+const InputWithTextFormat = props => (
+  <WithTextFormat {...props} render={renderProps => <input {...renderProps} />} />
+);
+
+InputWithTextFormat.propTypes = {
+  textFormatPattern: Types.string.isRequired,
+  onChange: Types.func.isRequired,
+};
+
+export default InputWithTextFormat;
```
StevePole: I'm a bit late in the day, but I think `InputWithDisplayFormat` would be a better name than `InputWithTextFormat`, similarly the prop would be better as `displayPattern` rather than `textFormatPattern`.  I think this just makes it clearer that this is _for display_ and not actually formatting the value.  It's also better distinguished from a validation pattern, which we also have at Field level.

src/select/Select.js
```diff
@@ -24,7 +24,9 @@ function actionableOption(option) {
 function stopPropagation(event) {
   event.stopPropagation();
   event.preventDefault();
-  event.nativeEvent.stopImmediatePropagation();
+  if (event.nativeEvent && event.nativeEvent.stopImmediatePropagation) {
+    event.nativeEvent.stopImmediatePropagation();
```
StevePole: Could benefit from a comment, why is this necessary?

src/inputWithTextFormat/InputWithTextFormat.js
StevePole: We also call it `displayFormat` in the V3 API. https://transferwise.atlassian.net/wiki/spaces/DF/pages/909346252/Presentation+options

docs/utils/generateInput/generatInput.js
StevePole: Why do we need separate handlers for checkbox changes vs other changes?

docs/utils/generateInput/generatInput.js
```diff
@@ -33,12 +40,28 @@ const generateInput = (knobs, componentThis) => {
...
             id={state}
             label={label}
-            onChange={value => componentThis.setState({ [state]: value })}
+            onChange={handleOnChangeCheckbox}
             checked={componentThis.state[state]}
           />
         </div>
       );
 
+    case 'number':
```
StevePole: I mentioned this in another PR ( https://github.com/transferwise/components/pull/161 ) but also relevant here. What we've effectively done in this file, is implement a dynamic FormControl, a much neater solution would be to use the better one you've just built!  :-)

src/common/Sizes.js
StevePole: The 'sm' _values_ may be a little tricky to change, but if you're looking for clarity you could change the constant
```
{
  EXTRA_SMALL: 'xs',
  SMALL: 'sm'
  MEDIUM: 'md',
  ...
}
```

docs/utils/generateInput/generatInput.js
andreapiras: Yes it's a simplified version. Can we achieve that by iterations since atm we don't have form control ready?

src/select/Select.js
pterblgh: Using `<Select>` in `<FormControl>` made the `<Select>` component work crazy and it triggered re-rendering hundred times within seconds which triggered React's error message about unnecessary changes. 

Maybe worth for others as well to check, possibly I was missing out something during my investigations.

src/common/unique-id.js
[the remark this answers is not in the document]
andreapiras: It feels a bit overkilled at the moment since the only usage is to generate the id for the tootltip. I'd leave the way it is and add the library if we need it for bigger purposes

src/common/sizes.js
```diff
@@ -0,0 +1,6 @@
+export const SIZES = {
```
dozortsev: We seem tend to use CamelCase for name of constants, should `Sizes` just like you created `FormControlType`

src/formControl/FormControl.js
```diff
@@ -0,0 +1,403 @@
...
```
dozortsev: shouldn't we add `id` to this input?

docs/utils/generateInput/generatInput.js
StevePole: Wouldn't it make this PR smaller though?  You wouldn't need this new code?

src/select/Select.js
StevePole: I meant a comment in the code :-).  Just so someone doesn't come along and delete something that looks redundant but is actually important.

src/withDisplayFormat/WithDisplayFormat.state.spec.js
```diff
@@ -1,45 +1,39 @@
...
-
-describe('InputWithTextFormat', () => {
+describe('WithDisplayFormat', () => {
   it('should return null if props change to the same value', () => {
-    const nextProps = { pattern: '**-**' };
+    const nextProps = { displayPattern: '**-**' };
     const prevState = {
       value: '11/11',
-      prevPattern: '**-**',
+      prevTextFormatPattern: '**-**',
```
StevePole: Weird that this one (and a few following) were not renamed?

PR #109, opened 2019-02-18
prefers map to forEach with new array

dey-dey:


taaviaasver:
Please bump version as well

PR #110, opened 2019-02-18
feat: update React to 16.8.2

pterblgh:
This will enable us writing components using Hooks API.


matallo:
dug out until realizing `getDerivedStateFromProps` is called after `setState` which is the expected behaviour (but breaks the component), this should be rearranged to prevent

pterblgh:
Until enzyme does not provide support for Hooks, I think we should leave this PR open and re-visit the topic once enzyme has provided alternatives to write tests for components using Hooks.

https://github.com/airbnb/enzyme/issues/2011

pterblgh:
Closing this as #175 has updated the React version.

PR #111, opened 2019-02-27
[WIP] Add button component 

domhorvathtw:
## Initial Button component

image
image
image
 
---

### Following design specs
https://transferwise.atlassian.net/wiki/spaces/DS/pages/859604203/Buttons

### Variants
- Primary
- Send
- Default
- Error

### Sates
- Default
- Loading 
- Disabled

### Different sizes 
- Extra small 
- Small 
- Medium (default) 
- Large

### Block option
Added isBlock flag

### Colours
Colour variables should get updated in bootstrap variables
https://github.com/transferwise/bootstrap/blob/master/less/variables.less

**Get new brand colour codes here:** 
https://app.goabstract.com/share/38662c17-b93d-454a-8a45-ea53b92e58f6

### Button layout variants
Circle, Dropdown and Split Button variants will be implemented in separate PR.


StevePole:
Couple quick things:

I think we need btn-default (blue outline, solid blue on hover/focus)
image

When block level, it would be nice if the loading indicator didn't cause the text to jump, there should generally be enough space, so I would fix the loader to the side.
image
Like the submit button does here:
image

Can't imagine when we'd use it, but the loader looks wrong on xs buttons, you could shrink it if it's simple to do
image

docs/ButtonDocs.js
```diff
@@ -0,0 +1,135 @@
+import React, { Component } from 'react';
+import { Button, Select, Checkbox } from '../src';
+
+const SIZES = ['xs', 'sm', 'md', 'lg'];
+const VARIANTS = ['primary', 'success', 'danger', 'warning'];
```
StevePole: We could have a conversation about whether these are the right names.  We don't need to be tied to the bootstrap class names, and the designers may prefer not to be.

docs/ButtonDocs.js
```diff
@@ -0,0 +1,135 @@
...
+const STATES = ['default', 'loading', 'disabled'];
+
+export default class ButtonDocs extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      label: 'Send money',
+      stateSelection: { value: 'default', label: 'default' },
+      variant: 'success',
```
matallo: we can probably use a map for variants, then set this to `VARIANTS.success` (this is just personal preference)

we're already exporting them in the component https://github.com/transferwise/components/pull/111/files#diff-b6f3260051f86253aa52309b2813c521R5

docs/ButtonDocs.js
```diff
@@ -0,0 +1,135 @@
...
+      size: 'md',
```
matallo: same comment than above for other variables like this

docs/ButtonDocs.js
domhorvathtw: Agree, makes sense

src/button/Button.js
```diff
@@ -0,0 +1,56 @@
...
+    [`btn-loading`]: isLoading,
+    [`btn-${variant}`]: variant,
+    [`btn-block`]: block,
+  });
+
+  return (
+    <button
+      type="button"
+      className={classes}
+      onClick={isDisabled ? null : e => onClick(e)}
```
matallo: 🤔 why not using always the `e` and `e.preventDefault()` when disabled?

docs/ButtonDocs.js
domhorvathtw: Totally agree, I personally think these are the use cases when Typescript and string enums are way to go.

src/button/Button.js
domhorvathtw: Thanks for pointing it out! Actually this whole logic is unnecessary. When a button HTML element has a `disabled` attribute, the user can't interact with the element. Removed it, and just used the onClick handle callback.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Also, testing it is unnecessary, since it's a browser feature, and testing if the disabled attribute is on the element when the prop is set, would be testing implement details, moreover, testing React, which is also unnecessary.

PR #113, opened 2019-02-28
Process indicator

domhorvathtw:
### Added process indicator component

image
image
image

**Props**
- size
- status

**Supported sizes** 
- xs
- sm
- xl

**Supported statuses** 
- processing (default)
- succeeded
- failed
- hidden

Transitions between states could be improved with CSS animations, keyframes.


docs/ProcessIndicatorDocs.js
```diff
@@ -0,0 +1,75 @@
+import React, { Component } from 'react';
+import { ProcessIndicator, Select, Radio } from '../src';
+import {
+  processIndicatorSizes,
+  processIndicatorStatuses,
+} from '../src/processIndicator/ProcessIndicator';
+
+export default class ProcessIndicatorDocs extends Component {
```
andreapiras: Please use PureComponents where possible

docs/ProcessIndicatorDocs.js
```diff
@@ -0,0 +1,75 @@
...
+  constructor(props) {
```
andreapiras: constructor is redundant here please use ```state = {
   status: 'processing',
      size: 'sm',
  };```

docs/ProcessIndicatorDocs.js
```diff
@@ -0,0 +1,75 @@
+import React, { Component } from 'react';
+import { ProcessIndicator, Select, Radio } from '../src';
+import {
+  processIndicatorSizes,
+  processIndicatorStatuses,
+} from '../src/processIndicator/ProcessIndicator';
```
andreapiras: If we use an index file this can be reduced to import {
  processIndicatorSizes,
  processIndicatorStatuses,
} from '../src/processIndicator';

docs/ProcessIndicatorDocs.js
```diff
@@ -0,0 +1,75 @@
...
+          <div className="row">
+            <div className="col-md-6">
+              <h2>Process Indicator</h2>
+              <p>
+                A component for conveying that a process is occuring, and for communicating a
+                successful or unsuccessful result
+              </p>
+            </div>
+            <div className="col-md-6">
+              <ProcessIndicator status={this.state.status} size={this.state.size} />
```
andreapiras: status and size can be deconstruct before render so we don't have to repeat this.state everywhere :)

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,35 @@
...
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+export const processIndicatorSizes = ['xs', 'sm', 'xl'];
+export const processIndicatorStatuses = ['processing', 'failed', 'succeeded', 'hidden'];
+const radius = { xs: 11, sm: 22, xl: 61 };
+
+const ProcessIndicator = ({ status, size }) => {
+  const classes = classNames('process', {
+    [`process-danger`]: status === 'failed',
```
andreapiras: I like this syntax 👍

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,35 @@
...
+    [`process-stopped`]: status === 'hidden',
+    [`process-success`]: status === 'succeeded',
+    [`process-${size}`]: size,
```
andreapiras: size is always defined as it si required so maybe this can be added at the top

src/processIndicator/ProcessIndicator.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    expect(wrapper.find('.process-sm')).toHaveLength(1);
+
+    expect(wrapper.find('.process-lg')).toHaveLength(0);
+    expect(wrapper.find('.process-xs')).toHaveLength(0);
+    expect(wrapper.find('.process-danger')).toHaveLength(0);
+    expect(wrapper.find('.process-stopped')).toHaveLength(0);
+    expect(wrapper.find('.process-success')).toHaveLength(0);
+  });
+
+  it('properly updates size and status classes on prop change', () => {
```
andreapiras: Don't we need the tests for the other status too?

andreapiras:
Nice one, super fast code delivery!!

src/processIndicator/ProcessIndicator.spec.js
domhorvathtw: added tests for both size and status classes

domhorvathtw:
After experimenting with the CSS for the process indicator from tw-bootstrap, I realized that a pure CSS3 solution (without a JS timing control) is not possible due to the nature of the animation. The spinning / chasing CSS animation's keyframes cannot be updated to form a full circle, because the stroke-dasharray and stroke-dashoffset attributes of the svg `<circle>` never form full ring. Therefore the only way to do that is changing the animation CSS property to a new animation, from `chasing-circle` to `close-circle`, however that could happen during the animation any time => jump.

**There is two options to create this process indicator component without controlling timing in the JS with setInterval()**

1. If the spinning animation doesn't modify the stroke attributes, they could be controlled by CSS transition, by default it would have a length of 1/3 of the circle's circumference, and in succeeded and failed state it would have the the length of the circumference.
2. The other solution would also require adjustments to the current animation, so that the last keyframe would display a full circle. By having that as the end of the animation, we would have to change only the `animation-iteration-time` from `infinite` to `1` at the end of the animation.

The current solution implements the setInterval solution on when the component mounts, then on every prop update it "waits" until the end of the animation to update the component's state and re-render the process indicator.

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+export const processIndicatorSizes = ['xs', 'sm', 'xl'];
+export const processIndicatorStatuses = ['processing', 'failed', 'succeeded', 'hidden'];
+const radius = { xs: 11, sm: 22, xl: 61 };
+const ANIMATION_DURATION_IN_MS = 1500;
+
+class ProcessIndicator extends Component {
```
andreapiras: Please use PureComponent

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+export const processIndicatorSizes = ['xs', 'sm', 'xl'];
+export const processIndicatorStatuses = ['processing', 'failed', 'succeeded', 'hidden'];
+const radius = { xs: 11, sm: 22, xl: 61 };
+const ANIMATION_DURATION_IN_MS = 1500;
```
andreapiras: should this maybe be a prop? and have a default of 1500?

domhorvathtw: I don't think it should be, it's hardcoded in the CSS animation, also can't think of a use-case when someone would want to slow down and speed up the spinning animation.

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,87 @@
...
+
+class ProcessIndicator extends Component {
+  static propTypes = {
+    status: Types.oneOf(processIndicatorStatuses).isRequired,
+    size: Types.oneOf(processIndicatorSizes).isRequired,
+  };
+
+  constructor(props) {
```
andreapiras: constructor is redundant

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,87 @@
...
+      size: 'sm',
+    };
+  }
+
+  /**
+   * Create interval for animation duration (1500ms)
+   * Update state only at the end of every interval
+   * (end of animation loop) if props changed before end of animation
+   */
+  componentDidMount() {
```
andreapiras: You may want to change this to getDerivedStateFromProps

src/processIndicator/ProcessIndicator.js
domhorvathtw: I'm using the shouldComponentUpdate lifecycle hook with custom logic for re-renders, the pureComponent would implement a shallow comparison, I think this change is not needed here

src/processIndicator/ProcessIndicator.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+export const processIndicatorSizes = ['xs', 'sm', 'xl'];
+export const processIndicatorStatuses = ['processing', 'failed', 'succeeded', 'hidden'];
+const radius = { xs: 11, sm: 22, xl: 61 };
```
andreapiras: should this maybe be a prop? and have a default of { xs: 11, sm: 22, xl: 61 }

src/processIndicator/ProcessIndicator.spec.js
```diff
@@ -0,0 +1,88 @@
...
+  afterEach(() => {
+    jest.useRealTimers();
+    jest.clearAllMocks();
+    jest.clearAllTimers();
+  });
+
+  it('has right classes when rendered with provided props', () => {
+    expect(wrapper.find('.process')).toHaveLength(1);
+    expect(wrapper.find('.process-sm')).toHaveLength(1);
+
```
andreapiras: extra row

src/processIndicator/ProcessIndicator.js
domhorvathtw: getDerivedStateFromProps runs in the update phase of the component as well (when props get updated),we should stick to componentDidMount, since we don't want to reset the interval every time props change. It's a mechanism to time updates for the end of every animation.

http://projects.wojtekmaj.pl/react-lifecycle-methods-diagram/

src/processIndicator/ProcessIndicator.js
domhorvathtw: these values are also linked to CSS, they should stay constant

PR #114, opened 2019-03-04
Bump formatting version

codewhisperer:
Formatting money strings with an invalid locale doesn't crash anymore, so the check in the demo was removed.


codewhisperer:
@Tankenstein Fixed

src/moneyInput/currencyFormatting.js
```diff
@@ -39,6 +39,17 @@ function isNumberLocaleSupported() {
   return numberString === '1,234';
 }
 
+function getValidLocale(locale) {
+  try {
+    const noUnderscoreLocale = locale.replace(/_/, '-');
+
+    Intl.NumberFormat(noUnderscoreLocale);
```
taaviaasver: Just started testing this.. are you sure its OK?

```
Intl.NumberFormat('xX');  // -> OK, but clearly invalid
```
while:
```
Intl.NumberFormat.supportedLocalesOf('xX') // -> empty array
```

Maybe we should switch to `supportedLocalesOf` check..

Issue #99, opened 2019-01-21
Modal can't be open by default


matallo:

czebe:
I encountered this issue, here's a fix that hopefully solves the problem:
https://github.com/transferwise/components/pull/119

PR #119, opened 2019-03-18
Allow modal to animate in when attached to DOM

czebe:
When rendering the modal conditionally immediately in the `open` state, the `in` classes were not applied, hence the modal exists in the DOM, but is invisible.

- Added `appear`:
> Normally a component is not transitioned if it is shown when the <Transition> component mounts. If you want to transition on the first mount set  appear to true, and the component will transition in as soon as the <Transition> mounts.

https://reactcommunity.org/react-transition-group/transition#Transition-prop-appear


andreapiras:
I think this is related to https://github.com/transferwise/components/issues/99 where modal couldn't be rendered as opend by default

PR #122, opened 2019-03-26
Np 33 process indicator callback

andreapiras:


src/processIndicator/ProcessIndicator.js
```diff
@@ -35,7 +42,10 @@ class ProcessIndicator extends Component {
       const sizeFromProps = this.props.size;
 
       if (statusFromState !== statusFromProps || sizeFromState !== sizeFromProps) {
-        this.setState({ status: statusFromProps, size: sizeFromProps });
+        this.setState(
+          { status: statusFromProps, size: sizeFromProps },
+          onAnimationCompleted && onAnimationCompleted(statusFromProps),
```
domhorvathtw: Calling the onAnimationCompleted callback here will run at the start of the ending animation.
The **succeded** and **failed** animations both take **1500ms**, just like the processing state animation loop.
It has to be wrapped in a setTimout call, so the component will "notify" the parent component that passed the `onAnimationCompleted` callback when it actually completed the animation. (e.g. it trurns green with the check icon or turns red with the cross)

PR #124, opened 2019-03-27
Definition list

OFFLlNE:
definitionListDemo

Moving the Definition list component from styleguide to Components by mostly migrating existing code to React

Work started by Peter now being finished.


src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+              return null;
+            }
+
+            return (
+              <div
+                className={!narrow && layout === 'vertical' ? 'col-xs-12 col-sm-6' : null}
+                // eslint-disable-next-line react/no-array-index-key
+                key={index}
+              >
+                {!field.group && (
```
StevePole: Surprised to see field.group here, does it exist in the angular one?  I thought we were using a version of the code from 'RequirementsService' to update legacy JSON.

src/definitionList/utils/text-format.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-plusplus */
+
+const formatUsingPattern = (value = '', pattern) => {
```
StevePole: not sure text format should be a child of definitionList, in fact I would guess it has already been moved to the formatting repo?

src/definitionList/utils/requirements.js
```diff
@@ -0,0 +1,483 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-param-reassign */
+
+const prepRequirements = alternatives => {
```
StevePole: This will be used by fieldsets as well as definition lists, so the file should not be a child of definitionList.

src/definitionList/DefinitionList.js
OFFLlNE: I assume here Peter, just moved it exactly as it is in styleguide.
So yea, it seems to be in angular one atm at least

https://github.com/transferwise/styleguide-components/blob/master/src/forms/definition-list/definition-list.html#L13

src/definitionList/utils/requirements.js
StevePole: In fact, this code may already exist somewhere in this repo.

src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+        {title && (
+          <div className="m-t-1">
+            <h4 className="page-header p-t-3">{title}</h4>
+          </div>
+        )}
+        <div className={!narrow && layout === 'vertical' ? 'row row-equal-height' : null}>
+          {Object.keys(fields).map((key, index) => {
+            const field = fields[key];
+
+            if (!(model[key] && !field.hidden)) {
```
pterblgh: Feel free to discard this comment because it's a style/taste of coding comment, but wouldn't this line be more readable with the following expression?
```js
if(!model[key] || field.hidden) {
  return null;
}
```

src/definitionList/DefinitionList.js
pterblgh: What @OFFLlNE says, I've just moved the existing interface and implementation from the AngularJs styleguide and converted it into a React one.

src/definitionList/utils/text-format.js
OFFLlNE: There does not seem to be `pattern` formatting in formatting repo. If I am looking correctly.
But something worth adding there actually - I might look into that today

src/definitionList/DefinitionList.js
OFFLlNE: I agree with this one :) less `!`s feels nicer

src/definitionList/utils/text-format.js
pterblgh: AFAIK the formatting library does not support only text formatting at the moment. I think this formatting logic is fine here in this library, I don't know any use-cases right now where we would need some kind of model driven text formatting - in this case the Dynamic Forms model provide the model.

If other component would require this kind of text formatting within this library we can move the code to a shared folder or create a dedicated component for this purpose. If another completely different lib would need this same behaviour than I would move it to formatting - but until this is not the case, I would let this code here.

src/definitionList/utils/requirements.js
pterblgh: Has any Dynamic Form consuming component moved to this repo? If not, I don't think there is any code here which handles requirements. 🤔

docs/DefinitionListDocs.js
```diff
@@ -0,0 +1,181 @@
+import React, { Component } from 'react';
+import { DefinitionList } from '../src';
+import Select from '../src/select';
+
+const LAYOUTS = ['vertical', 'horizontal', 'justified'];
```
andreapiras: Would make more sense to rename those as one-column,two-colum,one-column-justified

OFFLlNE: @StevePole WDYT? Since current naming is taken directly from styleguide?
Personally would favor the suggested change to reflect better :)

pterblgh: `one-column`,`two-columns` etc ties the abstraction to the visual representation. 

For example someone can easily refactor the component to a version which measures the innerWidth of the enclosing component and display 2 or 3 columns based on that value. In that case the naming convention would be inconsistent with the visual representation.

If you don't see this is something we should take into consideration, feel free to discard this comment. :)

src/definitionList/DefinitionList.js
StevePole: Ok, so my fault :-)

I don't think it would ever be necessary, the prepFields/prepRequirements code strips out field groups if they exist.

src/definitionList/utils/text-format.js
StevePole: This code covers formatting of input controls as well as pure presentation of values, hence the 'AfterCursor' stuff.  It's used by twField to format text in those controls.

docs/DefinitionListDocs.js
```diff
@@ -0,0 +1,181 @@
...
+
+export default class DefinitionListDocs extends Component {
```
andreapiras: Can we use PureComponent here?

docs/DefinitionListDocs.js
```diff
@@ -0,0 +1,181 @@
...
+  constructor(props) {
```
andreapiras: constructor is redundant here

src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+    title: Types.string,
+    narrow: Types.bool,
+    layout: Types.string,
+  };
+
+  static defaultProps = {
+    locale: 'en-GB',
+    title: null,
+    narrow: false,
+    layout: 'vertical',
```
andreapiras: maybe it would be good to define those layout in an array so if we'll ever change them it will be easier

src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  state = {
+    fields: this.props.fields,
+  };
+
+  componentWillMount() {
+    this.setState({
```
andreapiras: image

src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+      fields: prepFields(this.state.fields, this.state.model),
+    });
+  }
+
+  getClassNameForLine() {
```
andreapiras: This canbe simplified as ```getClassNameForLine = () => classNames({
      'dl-horizontal-item': this.props.layout === 'justified',
      'dl-horizontal': this.props.layout === 'horizontal',
    });```

andreapiras: This can be simplified as 
` getClassNameForLine = () => classNames({
      'dl-horizontal-item': this.props.layout === 'justified',
      'dl-horizontal': this.props.layout === 'horizontal',
    });`

src/definitionList/DefinitionListItem.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import Types from 'prop-types';
+import { formatDate, formatNumber } from '@transferwise/formatting';
+import { formatUsingPattern } from './utils/text-format';
+
+/**
```
andreapiras: I'm a big fan of JsDocs but I don't think we are using it in this project, maybe is worth checking in the guild if we can introduce it or if we have to keep consistency with what's in place already

src/definitionList/DefinitionListItem.js
```diff
@@ -0,0 +1,64 @@
...
+  field: Types.shape({
+    control: Types.string.isRequired,
+    displayFormat: Types.string,
+    refreshRequirementsOnChange: Types.bool,
+    title: Types.string.isRequired,
+    type: Types.string.isRequired,
+    width: Types.string,
+  }).isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  value: Types.any.isRequired,
```
andreapiras: Can we not have a list of all the different supported types?

src/definitionList/DefinitionListItem.spec.js
```diff
@@ -0,0 +1,174 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { formatNumber } from '@transferwise/formatting';
+import DefinitionListItem from './DefinitionListItem';
+
+describe('DefinitionListItem', () => {
+  let props;
+
```
andreapiras: weird spacing

src/definitionList/DefinitionListItem.spec.js
```diff
@@ -0,0 +1,174 @@
...
+          control: 'text',
+          displayFormat: '** - **',
+          refreshRequirementsOnChange: true,
+          title: 'Text',
+          type: 'string',
+          width: 'md',
+        },
+      };
+      component = shallow(<DefinitionListItem {...props} />);
+      span = component.find('span');
```
andreapiras: (Optional) This is a bit generic...It doesn't give any idea of what you are looking for. Can we maybe give it a className?

docs/DefinitionListDocs.js
OFFLlNE: And another reason why it cannot be two-columns - on smaller width it will go back to 1 column :D

docs/DefinitionListDocs.js
OFFLlNE: Don't we need that for state?

src/definitionList/utils/requirements.js
```diff
@@ -0,0 +1,483 @@
...
+  );
+};
+
+const fetchValuesAsync = (field, postData) => {
+  const { url } = field.valuesAsync;
+  const options = { method: field.valuesAsync.method || 'GET' };
+  if (postData) {
+    options.body = JSON.stringify(postData);
+  }
+  return fetch(url, options)
```
andreapiras: how do we handle eventual errors in the fetch? do we need a catch statement?

docs/DefinitionListDocs.js
OFFLlNE: Seems to work fine, so I guess, yes 😅

src/definitionList/DefinitionList.js
OFFLlNE: I think we do not need that for now. If in the future we would add more layouts, maybe worth doing that then, but right know I think the `getClassNameForLine` is already a good way to make it more easier in the future

src/definitionList/DefinitionList.js
OFFLlNE: I personally like return more, but I believe fat arrow is more conventional use. Will use your suggestion ;)

src/definitionList/utils/requirements.js
```diff
@@ -0,0 +1,483 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-param-reassign */
+
+const prepRequirements = alternatives => {
+  if (!alternatives || !alternatives.length) {
+    return [];
+  }
+
+  const preppedAlternatives = copyOf(alternatives);
```
andreapiras: This can be replaced with const preppedAlternatives = {...alternatives};

src/definitionList/DefinitionListItem.js
OFFLlNE: @pterblgh Any specific reason of adding that? Maybe you can comment on this :)

src/definitionList/utils/requirements.js
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  const values = field.enum || field.values;
+  if (values) {
+    return values.length > 3 ? 'select' : 'radio';
+  }
+  return 'select';
+};
+
+const copyOf = obj => JSON.parse(JSON.stringify(obj));
```
andreapiras: (Optional) this can be simplified as const copyOf = obj => ({...obj});  if we don't need a deep copy

src/definitionList/DefinitionListItem.spec.js
OFFLlNE: Can you give an example what do you mean?

src/definitionList/DefinitionList.js
```diff
@@ -0,0 +1,100 @@
...
+          {Object.keys(fields).map((key, index) => {
+            const field = fields[key];
+
+            if (!(model[key] && !field.hidden)) {
+              return null;
+            }
+
+            return (
+              <div
+                className={!narrow && layout === 'vertical' ? 'col-xs-12 col-sm-6' : null}
```
andreapiras: (Optional) I find this more readable `{classNames({
          'col-xs-12 col-sm-6': !narrow && layout === 'vertical',
        })}`

src/definitionList/utils/requirements.js
andreapiras: I've implemented something similar
`fetch(`${httpOptions.url}`, {
    ...httpOptions,
    body: data,
  })
    .then(response => {
      if (!response.ok) {
        const errorMessage = { status: response.status, statusText: response.statusText };
        throw errorMessage;
      }
      return response;
    })
    .catch(e => {
      throw e;
    });`

src/definitionList/utils/text-format.js
```diff
@@ -0,0 +1,48 @@
...
+  if (typeof pattern !== 'string') {
+    return value;
+  }
+
+  if (pattern.indexOf('||') > 0) {
+    // eslint-disable-next-line no-param-reassign
```
andreapiras: Maybe we should create another variable :)

src/definitionList/DefinitionListItem.js
pterblgh: In some modern IDEs that understands JSDoc you can have some nice intellisense and you can see the proper types for the arguments of a function. That's the only reason I'm using them - also in the future if we're considering using TypeScript for type checking, because the tool understands JSDoc annotations and can just type check it without any transpilation.

src/definitionList/DefinitionList.js
OFFLlNE: Tried DidMount with you - did not work as intended.
Added UNSAFE to function name

src/definitionList/DefinitionListItem.js
OFFLlNE: I would say that it does not hurt to have that :)

src/definitionList/DefinitionListItem.js
OFFLlNE: boolean, date, string, int are some of them - might be more.
I can add those as a comment if you think that would be a good thing to do

src/definitionList/utils/text-format.js
OFFLlNE: Not sure we really can do that nicely - because not all patterns go into this `if block` and because of that I believe this is the cleanest way to do it.

src/definitionList/utils/requirements.js
OFFLlNE: I would keep it as it is just for readability :)

src/definitionList/utils/requirements.js
OFFLlNE: So just adding
```
.catch(e => {
      throw e;
    });
```
is sufficient in this case I would assume

src/definitionList/DefinitionListItem.spec.js
OFFLlNE: I changed the variable to `definitionListItem`.

src/definitionList/utils/requirements.js
OFFLlNE: I believe the catching is being done above by calling out this function
```
fetchValuesAsync(field, postData).catch(() =>
    // Retry once on failure
    fetchValuesAsync(field, postData),
  );
};
```

src/definitionList/DefinitionList.js
andreapiras: Did we not talk about change this in the constructor since this.state.model is undefined anyway...

src/definitionList/DefinitionListItem.js
andreapiras: I'd like that too , just maybe it's worth it to talk about it in the FEG so we are all on the same page

src/definitionList/DefinitionListItem.js
andreapiras: Nah let's leave it as it is

src/definitionList/utils/text-format.js
andreapiras: The linter here is complaining about the re-assignment of the function parameter pattern. So all we need to do is to assign this at line 7 let newPattern = pattern;  and work with that rather than with pattern directly

docs/DefinitionListDocs.js
```diff
@@ -0,0 +1,177 @@
...
+                </pre>
+                {/* eslint-enable react/jsx-indent */}
+              </div>
+              <label htmlFor="definition-list-layout">Layout</label>
+              <Select
+                id="definition-list-layout"
+                selected={
+                  this.state.layout
+                    ? { value: this.state.layout, label: this.state.layout }
+                    : 'vertical'
```
andreapiras: LAYOUTS[0]

docs/DefinitionListDocs.js
```diff
@@ -0,0 +1,177 @@
...
+                }
+                options={LAYOUTS.map(s => ({ value: s, label: s }))}
+                onChange={selection =>
+                  this.setState({ layout: selection ? selection.value : 'vertical' })
```
andreapiras: LAYOUTS[0]

src/definitionList/utils/text-format.js
OFFLlNE: I did that yea now. But it is not always a `new` pattern :D That is why I did not want to do that before :D

src/definitionList/utils/text-format.js
```diff
@@ -0,0 +1,49 @@
...
+
+const countSeparatorsAfterCursor = (newPattern, position) => {
+  let separators = 0;
+  while (positionIsSeparator(newPattern, position + separators)) {
+    separators++;
+  }
+  return separators;
+};
+
+function positionIsSeparator(newPattern, position) {
```
andreapiras: Optional 
`
const positionIsSeparator = (newPattern, position) => newPattern[position] && newPattern[position] !== '*';
`

PR #127, opened 2019-04-03
Moved docs into components folder

andreapiras:


docs/Docs.js
```diff
@@ -1,49 +1,15 @@
...
-import CheckboxOptionDocs from './CheckboxOptionDocs';
-import NavigationOptionDocs from './NavigationOptionDocs';
-import DateLookupDocs from './DateLookupDocs';
-import PhoneNumberInputDocs from './PhoneNumberInputDocs';
-import DateInputDocs from './DateInputDocs';
-import InstructionListDocs from './InstructionListDocs';
-import ProcessIndicatorDocs from './ProcessIndicatorDocs';
+
+const req = require.context('../src', true, /\.docs\.js$/);
+const loadDocs = () => req.keys().map(filename => req(filename));
```
skywritergr: This is cleaner and I like it. I would just add a comment block above to briefly describe what's going on. For future dev generations.

StevePole:
Nice 👍

src/checkbox/Checkbox.docs.js
```diff
@@ -1,7 +1,7 @@
-import React, { Component } from 'react';
-import { Checkbox } from '../src';
+import React, { PureComponent } from 'react';
+import { Checkbox } from '../';
```
taaviaasver: Maybe replace with `import Checkbox from '.';`?
(same thing seems to apply to all other `.docs.js` files as well)

Or when you import multiple components, then `'..'` -> I think we can omit the trailing `/`

PR #128, opened 2019-04-04
Single comp dev

andreapiras:


StevePole:
Could you screenshot?

andreapiras:
singlepage
@StevePole

StevePole:
Why does checkbox appear on both pages?  Was expecting to see more of a navigation on the first page, then the actual component when you click in.

andreapiras:
> Why does checkbox appear on both pages? Was expecting to see more of a navigation on the first page, then the actual component when you click in.

To be honest I'm not at that stage yet. This PR is more about adding the route generated automatically and to create a individual page for each component. All the details about how the navigation shoul behave is still tbd and open to discussion

StevePole:
Ok, I think it's fine to do it incrementally, but I may avoid including the links to the new routes until we're ready.  So the components will be available on the main page, or on a route.  In a follow up we add navigation, and remove the docs from the main page?

andreapiras:
Sure the link can be removed and leave the route available. I'm working on a another branch where I'm actually adding a sidebar and trying to set up a nice navigation. This PR is more of a show case where we can evaluate the code to achieve this setup

PR #129, opened 2019-04-04
Add definitionList

OFFLlNE:
Everything started with this PR: https://github.com/transferwise/components/pull/124
Feel free to check up there for previous reviews and pictures how the thing looks like

Apparently on this new branch CircleCI is finally happy although exact same things as in #124.


src/definitionList/utils/text-format.js
```diff
@@ -0,0 +1,48 @@
...
+
+  while (charactersToAllocate) {
+    if (positionIsSeparator(newPattern, position)) {
+      newValue += newPattern[position];
+      separators++;
+    } else {
+      newValue += value[position - separators];
+      charactersToAllocate--;
+    }
+    position++;
```
codewhisperer: how about `postition += 1`?

src/definitionList/utils/text-format.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-plusplus */
```
codewhisperer: I don't think you need this

codewhisperer:
Noice!

PR #131, opened 2019-04-09
Field component

dozortsev:
## Demo

https://transferwise.github.io/components/branch/feat-field/index.html#field ([Angular's component](http://transferwise.github.io/styleguide-components/requirements.html#field) just for comparison)

## Jira

https://transferwise.atlassian.net/browse/NP-13


src/common/validation/rule-validators/index.js
```diff
@@ -0,0 +1,47 @@
+function isValidRequired(value, isRequired) {
```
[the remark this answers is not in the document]
jacklesliewise: Where are these tests now?

src/field/Field.js
```diff
@@ -0,0 +1,264 @@
...
+    }
+    if (this.dirty) {
+      this.validateValue(event.target.value);
+    }
+  }
+
+  onChange(event) {
+    if (this.props.onChange) {
+      this.props.onChange(event);
+      this.dirty = true;
```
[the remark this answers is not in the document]
jacklesliewise: Is this still an issue? I don't see `dirty` in `Field` or `FormControl`

src/field/Field.js
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    if (this.props.errorMessage) {
+      this.setState({ customErrorMessage: null });
+    }
+
+    this.validateValue(event.target.value);
```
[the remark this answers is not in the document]
jacklesliewise: Same as this https://github.com/transferwise/components/pull/131#discussion_r318956434

src/common/validation/rule-validators/index.js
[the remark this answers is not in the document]
dozortsev: [here](https://github.com/transferwise/styleguide-components/blob/889e5b62c980739887e408cea85310e6ab8eaaa4/src/json-schema/validation/rule-validators/spec.js#L10) could you please copy them here as well

src/common/validation/type-validators/index.js
```diff
@@ -0,0 +1,20 @@
+function isString(value) {
```
dozortsev: please add tests for this as well, you can find examples which Steve implemented [here](https://github.com/transferwise/styleguide-components/blob/889e5b62c980739887e408cea85310e6ab8eaaa4/src/json-schema/validation/type-validators/spec.js)

src/common/validation/validation-failures/index.js
```diff
@@ -0,0 +1,144 @@
+import { isObject, isArray, isString, isNumber, isInteger, isBoolean } from '../type-validators';
```
dozortsev: [same](https://github.com/transferwise/styleguide-components/blob/889e5b62c980739887e408cea85310e6ab8eaaa4/src/json-schema/validation/validation-failures/spec.js) for these functions

src/field/Field.docs.js
```diff
@@ -0,0 +1,423 @@
...
+
+    const selectedComponent = FIELDS.find(field => field.formControl === control);
+
+    const controlIsNumber = control === FormControlType.NUMBER;
+
+    selectedComponent.fieldProps.field.step = controlIsNumber ? step : null;
+    selectedComponent.fieldProps.field.minimum = controlIsNumber ? minimum : null;
+    selectedComponent.fieldProps.field.maximum = controlIsNumber ? maximum : null;
+
+    selectedComponent.fieldProps.errorMessage = addErrorMessage ? 'Custom error message' : null;
```
dozortsev: we better to show actual component props rather than this..

src/common/validation/rule-validators/index.js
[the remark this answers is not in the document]
jacklesliewise: Added in 8e87153

src/common/validation/type-validators/index.js
jacklesliewise: Added in 8e87153

src/common/validation/validation-failures/index.js
jacklesliewise: Added in 8e87153

src/field/Field.docs.js
jacklesliewise: Added in 005e100

StevePole:
Latest version of the spec does this slightly differently

image

Now:
image

StevePole:
Can't disabled radio buttons (or readOnly):
image

StevePole:
Hidden isn't a field type, but a separate option like disabled/readOnly.

image

StevePole:
Error messages don't show up (warnings do)
image

StevePole:
If we have help information we need to disable browser autocomplete, otherwise it's hidden
image

https://github.com/transferwise/styleguide-components/blob/master/src/forms/form-control/form-control.html#L127

https://github.com/transferwise/styleguide-components/blob/master/src/forms/form-control/form-control.controller.js#L35 (there's a bug there relating to the help information structure).

StevePole:
Don't duplicate label in date lookup.  There is an extra parameter for inline labels, but we won't use it in this case.
image

StevePole:
ReadOnly checkboxes are not read only
image

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+      expect(component.find(FormControl)).toHaveLength(1);
+      expect(
+        component
+          .find(FormControl)
+          .first()
+          .props().type,
+      ).toBe(FormControlType.CHECKBOX);
+    });
+
+    it('should render a visible label', () => {
```
StevePole: 'should hide'?

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+        .find(FormControl)
+        .first()
+        .props();
+
+      expect(formControlProps.type).toBe(FormControlType.TEXT);
+      expect(formControlProps.value).toBe('supplied value');
+    });
+  });
+
+  describe('when given an error message', () => {
```
StevePole: Not sure these error message tests are sufficient as I had cases where I couldn't get an error message to show.

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+    });
+  });
+
+  describe('when given a help message', () => {
+    beforeEach(() => {
+      props = {
+        field: {
+          label: 'Control with help message',
+          type: FieldTypes.STRING,
+          helpText: 'Some useful hint',
```
StevePole: Change format to:
```
help: {
  message: "...:
}
```

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+      it('should invoke on focus handler', () => {
+        expect(props.onFocus).toHaveBeenCalled();
+      });
+
+      it('should be in focused and not be in blur state', () => {
+        expect(component.state().focused).toBeTruthy();
+      });
+
+      it('should add info state', () => {
+        expect(component.find('.has-info')).toHaveLength(1);
```
StevePole: Can you try changing this to `.has-focus` (in both tests and implementation), I think it exists and is a more semantic option.

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+        expect(alertProps.children.props.children).toBe('Some useful hint');
+        expect(alertProps.type).toBe(AlertType.Info);
+        expect(alertProps.size).toBe(AlertSize.Small);
+        expect(alertProps.arrow).toBe(AlertArrowPosition.UpLeft);
+      });
+    });
+
+    describe('when control not on focus', () => {
+      it('should not add info state', () => {
+        expect(component.find('.has-info')).toHaveLength(0);
```
StevePole: Also `.has-focus`

src/common/requirements.js
```diff
@@ -1,40 +1,6 @@
 /* eslint-disable no-unused-vars */
 /* eslint-disable no-param-reassign */
 
-const prepRequirements = alternatives => {
```
StevePole: How can we get rid of this?

src/field/Field.js
```diff
@@ -0,0 +1,294 @@
...
+        onCancel: Types.func,
+        onFailure: Types.func,
+        onStart: Types.func,
+        onSuccess: Types.func,
+        psButtonText: Types.string,
+        psFailureText: Types.string,
+        psProcessingText: Types.string,
+        psSuccessText: Types.string,
+        size: Types.string,
+        usAccept: Types.oneOf(['*', 'image/*', 'application/*']),
```
StevePole: What are the `cs`,`ps` and `us` prefixes about?

StevePole:
Can we make some differentiation between the configuration controls and the component, found it a bit confusing.  Maybe just a `<hr /` would be enough to separate them.

image

src/field/Field.spec.js
jacklesliewise: This doesn't appear to exist when I change and run locally

src/field/Field.spec.js
jacklesliewise: See above

src/formControl/FormControl.js
```diff
@@ -406,6 +381,7 @@ export default class FormControl extends PureComponent {
         return <textarea {...textareaProps} />;
       }
 
+      case FormControlType.FILE:
```
StevePole: Why do we need this?  If we have legacy file types to support we should manage that through the 'prep' functions, and keep the implementation clean.

src/field/Field.js
```diff
@@ -0,0 +1,324 @@
...
+      hidden: Types.bool,
+      readOnly: Types.bool,
+      autoComplete: Types.bool,
+      placeholder: Types.string,
+      searchPlaceholder: Types.string,
+      control: Types.oneOf(Object.values(FormControlType)),
+      format: Types.oneOf(Object.values(FieldFormats)),
+      displayPattern: Types.string,
+      help: Types.shape({
+        text: Types.string,
```
StevePole: This should be 'message'

StevePole: Can you check that the 'prep' code will turn `helpText` into `help: {message: ...}`, helpList into help: list etc?  This happens on the angular version of that code, but not sure when it was forked.

src/field/Field.spec.js
```diff
@@ -0,0 +1,539 @@
...
+    it('should render the message', () => {
+      const alert = component.find(Alert);
+      expect(alert).toHaveLength(1);
+
+      const alertProps = alert.first().props();
+      expect(alertProps.children.props.children).toBe('Custom error');
+      expect(alertProps.type).toBe(AlertType.Error);
+      expect(alertProps.size).toBe(AlertSize.Small);
+      expect(alertProps.arrow).toBe(AlertArrowPosition.UpLeft);
+    });
```
StevePole: Can you add tests for:

- when the value changes
  - it should remove the error message
  - when a new error message is supplied
    - it should show the new error message
?

In actual fact, there is a test for removing the error below, but I think it would be better grouped here, with the additional test.

PR #132, opened 2019-04-12
Use touchstart for ios devices

KarlEST:
Currently you can't close popups and selectors on ios devices(iphone, ipad). This is because ios doesn't recognise `click` events on document(it's not clickable element so it's not fired there). This means we have to use `touchstart` for these cases. Reason why we won't use `touchstart` for all devices is because we don't want popover to disappear when you scroll(scrolling triggers `touchstart` but not `click`).


KarlEST:
> Looks good! Wouldn't the same thing be useful Android as well?

@codewhisperer Android works with `click` event and we prefer using that one. So when you scroll the page `Popover` stays open

src/popover/Popover.js
```diff
@@ -27,7 +28,11 @@ export default class Popover extends Component {
   };
 
   componentWillUnmount() {
-    document.removeEventListener('click', this.closePopoverOnOutsideClick, true);
+    document.removeEventListener(
+      isIosDevice() ? 'touchstart' : 'click',
```
mikknarusk: maybe we can extract this logic into a function?

src/select/Select.js
```diff
@@ -195,13 +196,21 @@ export default class Select extends Component {
         this.searchBox.focus();
       }
     });
-    document.addEventListener('click', this.handleDocumentClick, false);
+    document.addEventListener(
+      isIosDevice() ? 'touchstart' : 'click',
```
mikknarusk: and this

src/common/deviceDetection/deviceDetection.js
```diff
@@ -0,0 +1,7 @@
+export function isIosDevice() {
+  const sampleTouchDevices = ['iphone', 'ipad', 'ipod'];
+  const matchString = sampleTouchDevices.map(device => `(${device})`).join('|');
+  const regex = new RegExp(matchString, 'ig');
+
+  return navigator && !!navigator.userAgent.match(regex);
```
Tankenstein: check typeof navigator

PR #133, opened 2019-04-15
Input field

andreapiras:
**Why**
This input field originally implemented in [angular](http://transferwise.github.io/styleguide-components/index.html#text-format) was developed in order to assists customer while inserting specific data.

Input with text format is a component that formats text entered into controls on typing. The unformatted value is broadcasted on change

**Demo**
https://transferwise.github.io/components/branch/input-field/index.html


src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,240 @@
...
+        toDelete = 1;
+      }
+
+      newStack.splice(startPosition, toDelete);
+    }
+
+    return newStack;
+  };
+
+  handleCursor = unformattedValue => {
```
StevePole: handleCursor could have a more descriptive name

src/inputWithTextFormat/InputWithTextFormat.docs.js
```diff
@@ -0,0 +1,185 @@
...
+              {this.getDocs()}
+            </div>
+            <div className="col-md-6">
+              <div
+                className={classNames({
+                  ' m-b-4': true,
+                  'droppable-dropping': this.state.isDropping,
+                })}
+              >
+                <InputWithTextFormat
```
StevePole: Why InputWithTextFormat and not InputWithOnChange or InputWithPlaceholder.  I don't think TextFormat needs to be special.  Could we make it TextInput with an optional pattern parameter? If you don't supply the pattern it's just a regular TextInput?

src/common/textFormat/getPatternSymbolsWithPosition/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './getPatternSymbolsWithPosition';
```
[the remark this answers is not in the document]
andreapiras: The only advantage I can see is on import, using the index prevents you to specify the file in the. The only reason I can see of not putting the code in the index is that otherwise in your editor if multiple index are open it would be confusing having all the files named index. Apart from that I don't have a strong opinion about

src/inputWithTextFormat/InputWithTextFormat.docs.js
andreapiras: I'm up for changing the name. Regarding using this component when pattern is not provided, I think it wouldn't be very optimized and we can use a native input field if we don't need any customized behaviour

src/inputWithTextFormat/InputWithTextFormat.docs.js
```diff
@@ -0,0 +1,185 @@
...
+      }),
+    );
+    return state;
+  };
+
+  getDocs = () => {
+    const docs = Object.values(KNOBS)
+      .map(knob =>
+        knob
+          .reduce((acc, curr) => {
```
StevePole: Could we make a function with a descriptive name for this reducer?

src/common/textFormat/unformatWithPattern/unformatWithPattern.js
```diff
@@ -0,0 +1,16 @@
+const unformatWithPattern = (value, pattern) =>
+  value && value.length
+    ? value
+        .toString()
+        .split('')
+        .filter(
+          el =>
```
StevePole: Could we make a function with a descriptive name for this filter?

src/inputWithTextFormat/InputWithTextFormat.spec.js
```diff
@@ -0,0 +1,140 @@
...
+  { value: '1234', expectedValue: '12-34', pattern: '**-**-**' },
+  { value: '1234', expectedValue: '+(12) 34', pattern: '+(**) *' },
+];
+const props = {
+  placeholder: 'a placeholder',
+  pattern: '**-**',
+  className: 'form-control',
+  onChange: jest.fn(),
+};
+describe('InputWithTextFormat', () => {
```
StevePole: Line breaks :-)

src/common/textFormat/getNextGroupedSymbols/getNextGroupedSymbols.js
```diff
@@ -0,0 +1,36 @@
+/*eslint-disable */
+import getPatternSymbolsWithPosition from '../getPatternSymbolsWithPosition';
+
+const getNextGroupedSymbols = (selectionStart, pattern, direction = 'default') => {
```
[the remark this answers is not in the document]
StevePole: What is 'direction'?  I'd like to give more advice on better names for some of the stuff here, but I can't actually figure out what it does.

src/common/textFormat/getNextGroupedSymbols/getNextGroupedSymbols.js
```diff
@@ -0,0 +1,35 @@
...
+  const patternArray = getPatternSymbolsWithPosition(pattern);
+  let cursor = selectionStart;
+  let groupSize;
+
+  if (direction === 'default') {
+    const applicablePattern = patternArray.filter(symbol => symbol.index >= selectionStart);
```
StevePole: Perhaps the two implementations in this conditional ('default' and whatever the opposite of default is), could have named functions that describe what they're doing.

src/common/textFormat/getNextGroupedSymbols/getNextGroupedSymbols.spec.js
```diff
@@ -0,0 +1,26 @@
...
+  { selectionStart: 0, pattern: '**-**-**', expectedValue: 0, direction: 'default' },
+  { selectionStart: 2, pattern: '**-**-**', expectedValue: 1, direction: 'default' },
+  { selectionStart: 5, pattern: '**-**-**', expectedValue: 1, direction: 'default' },
+  { selectionStart: 0, pattern: '+(**)', expectedValue: 2, direction: 'default' },
+  { selectionStart: 0, pattern: '+(---***)', expectedValue: 5, direction: 'default' },
+  { selectionStart: 1, pattern: '+(---***)', expectedValue: 4, direction: 'default' },
+  // Reverse
+  { selectionStart: 1, pattern: '+(---***)', expectedValue: 1, direction: 'reverse' },
+  { selectionStart: 2, pattern: '+(---***)', expectedValue: 2, direction: 'reverse' },
+  { selectionStart: 3, pattern: '+(---***)', expectedValue: 3, direction: 'reverse' },
```
StevePole: Even using these I can't figure out what this function does.

StevePole: It is how many places to the right I have to go from the current position before I fine a non symbol character?  Or left if reversing position, and returning distance to start of string if none found?

src/inputWithTextFormat/InputWithTextFormat.docs.js
StevePole: I'm thinking about how this is going to be consumed in a dynamic form.  Particularly within the FormControl/Field components where we have a switch on the the data/control type to determine which component to use.  We'll now have to have an additional conditional statement to use two different components depending on whether the field has a display pattern, which feels a bit odd...
```
switch(field.type) {
  case 'string':
    if (field.displayFormat) {
      return <InputWithTextFormat ... >
    } else {
      return <input ... >
    }
```

andreapiras: I got what you mean but I wouldn't create a single component as if we don't need the textFormat functionality we should indeed use a native input

src/common/textFormat/getNextGroupedSymbols/getNextGroupedSymbols.spec.js
andreapiras: > It is how many places to the right I have to go from the current position before I fine a non symbol character? Or left if reversing position, and returning distance to start of string if none found?

correct :)

src/common/textFormat/getNextConsecutiveSymbols/getNextConsecutiveSymbols.js
```diff
@@ -0,0 +1,29 @@
+import getSymbolsInPatternWithPosition from '../getSymbolsInPatternWithPosition';
+
+const getNextConsecutiveSymbols = (selectionStart, pattern, direction = 'default') => {
```
StevePole: How about 'getDistanceToNextSymbol'?  getNextConsecuitveSymbols sounds like you're going to return a set of symbols that appear consecutively.

I would also change direction to be 'left' or 'right', reorder the parameter so that pattern is first (I'd say it's the primary argument), and rename selectionStart to be position.  So you have:
`const getDistanceToNextSymbol = (pattern, position, direction = 'right') => { ... }`

Or, even better, export two functions:
```
getDistanceToNextSymbol = (pattern, position) => { ... }
getDistanceToPreviousSymbol = (pattern, position) => { ... }
```

src/common/textFormat/getSymbolsInSelection/getSymbolsInSelection.js
```diff
@@ -0,0 +1,8 @@
+import { getSymbolsInPatternWithPosition } from '..';
+
+const getSymbolsInSelection = (selectionStart, selectionEnd, pattern) =>
```
StevePole: This doesn't return symbols!  How about:
````
const getCountOfSymbolsInSelection(pattern, selectionStart, selectionEnd)
```
?

src/common/textFormat/getCursorPosition/getCursorPosition.js
```diff
@@ -0,0 +1,34 @@
+import { getNextConsecutiveSymbols, getSymbolsInSelection } from '../';
+
+const getCursorPosition = (key, selectionStart, selectionEnd, pattern, pastedLength = 0) => {
```
StevePole: This one needs a bit more context.  Perhaps `getCursorPositionAfterKeystroke`?

src/inputWithTextFormat/utils/historyNavigator/historyNavigator.js
```diff
@@ -0,0 +1,36 @@
...
+    this.historyLimit = historyLimit;
+  }
+
+  reset = (newHistory = [''], historyLimit = 20) => {
+    this.history = newHistory;
+    this.currIndex = newHistory.length - 1 > 0 ? newHistory.length - 1 : 0;
+    this.historyLimit = historyLimit;
+  };
+
+  setHistory = element => {
```
StevePole: I don't think these two functions are gets and sets.  Set typically sets a whole value, get typically returns a whole value.  These are more like 'push'es and 'pop's, we're pushing something onto a stack, and popping something off.  Or rather than passing a 'action' as an argument, be more explicit with named functions for the possible arguments, I presume those are 'undo', 'redo', 'add'.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+    this.setState(
+      { value: formattedValue },
+      this.resetEvents(),
+      historyNavigator.add(formattedValue),
+      onChange(unformattedValue),
+    );
+  };
+
+  handleDelete = unformattedValue => {
+    const key = this.determineAction(unformattedValue);
```
StevePole: Could rename this variable to be 'action'?

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+      selectionEnd,
+      pattern,
+      pastedLength,
+    );
+
+    this.setCursorPosition(cursorPosition);
+  };
+
+  determineAction = unformattedValue => {
+    const { isPasting, isCutting, currentAction, value } = this.state;
```
StevePole: I find it odd that this function determines the action via state.  It would make more sense for this method to receive the event.  By the looks of it, you would also need to pass the current value, but I think that's fine.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+
+  isKeyAllowed = () => {
+    const { currentAction } = this.state;
+    const { pattern } = this.props;
+    const symbolsInPattern = pattern.split('').filter(el => el !== '*');
+
+    return symbolsInPattern.indexOf(currentAction) === -1;
+  };
+
+  handleOnChange = e => {
```
StevePole: Can you rename these variables to be `event`?  Truncating variable names hurts readability.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+      isCutting: false,
+      isPasting: false,
+      pastedLength: 0,
+    });
+  };
+
+  isKeyAllowed = () => {
+    const { currentAction } = this.state;
+    const { pattern } = this.props;
+    const symbolsInPattern = pattern.split('').filter(el => el !== '*');
```
StevePole: Rename `el`, I assume this is short for element, which is typically used to refer to HTML DOM elements, If I'm in the context of an iterator, I'd call it `item` or better, what it actually is, in this case `character`.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+      const { value, prevPattern, historyNavigator } = prevState;
+      const { pattern } = nextProps;
+      const unFormattedValue = unformatWithPattern(value, prevPattern);
+      historyNavigator.reset();
+
+      return {
+        prevPattern: pattern,
+        value: formatWithPattern(unFormattedValue, pattern),
+        isCutting: false,
+        isPasting: false,
```
StevePole: Ideally I would remove `isCutting` and `isPasting` from state, they are not long lived state, they are momentary things which we should be able to handle as they happen.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  setCursorPosition = select => {
+    setTimeout(() => {
+      this.input.current.setSelectionRange(select, select);
+      this.setState({ selectionStart: select, selectionEnd: select });
+    }, 0);
+  };
+
+  getKeyCombination = e => {
```
StevePole: It's weird that this method also returns the current Action.  You should merge this with `determineAction` (getCurrentAction would be a better name) and have one way of getting the action.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+
+    this.setState({
+      currentAction,
+      selectionStart,
+      selectionEnd,
+    });
+  };
+
+  handleOnPaste = e => {
+    let pastedChar = e.clipboardData.getData('Text').split('');
```
StevePole: What's a 'char'?  Presumably a 'character', but this variable does not contain a character.

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+    const { pattern } = this.props;
+    // Remove symbols from cut text.
+    pastedChar = pastedChar.filter(el => pattern.split('').indexOf(el) === -1);
```
StevePole: Could you achieve the same thing by using 'unformatWithPattern' and not turning this into an array?

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    this.setState({
+      currentAction,
+      selectionStart,
+      selectionEnd,
+    });
+  };
+
+  handleOnPaste = e => {
```
StevePole: We don't update history on paste or cut.  I wonder if history could be managed within the `setState` function to make sure it's always updated?

src/inputWithTextFormat/InputWithTextFormat.js
```diff
@@ -0,0 +1,244 @@
...
+    if (typeof triggerEvent.key === 'undefined') {
+      if (unformattedValue.length <= unformatWithPattern(value, pattern).length) {
+        return 'Backspace';
+      }
+    }
+
+    return triggerEvent.key;
+  };
+
+  detectUndoRedo = event => {
```
StevePole: I feel this would be better as an isRedo and isUndo

PR #134, opened 2019-04-23
Adjust DefinitionList

OFFLlNE:
### Visual changes are made only to the Justified layout

- Instead of `dl-horizontal-item` use `dl-horizontal`. By using that it will enable horizontal layout on `XS` and justified on other widths by adding `text-sm-right`
- If the API passes the element to be bold it will be made bold:
```
tagClassName: {
          h3: true,
        },
```
☝️ in the `field` makes that element bold if that is type `Text`

Was (SM - ...)            | Will (SM - ...)
:-------------------------:|:-------------------------:
  |  

Was (XS)            | Will (XS)
:-------------------------:|:-------------------------:
  |


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v4.18.3
+## Adjust DefinitionList
+On justified layout force text to the left, use dl-horizontal for smaller height between items
```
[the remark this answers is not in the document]
OFFLlNE: Good question actually.

I assume nothing will actually break if people have an extra property being passed in - maybe just some warnings of unused prop.
But since I do not know if anybody outside TransferWise managed to find this component and start using it in the last week, then I think the correct would be to have a major version bump indeed.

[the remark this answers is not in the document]
OFFLlNE: Because I initially knew that nobody in TW has the time yet to use that, but since this is public major is correct jap

src/definitionList/DefinitionList.spec.js
```diff
@@ -0,0 +1,157 @@
...
+            title: 'Date label',
+            type: 'date',
+          },
+        },
+      };
+      component = shallow(<DefinitionList {...props} />);
+    });
+
+    it('should have an element with col-sm-6 class', () => {
+      const rowElements = component.find('.col-sm-6');
```
andreapiras: uhmmm is it always the case? Should the page dictate the container width?

src/definitionList/DefinitionList.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    it('should not have an element with row-equal-height class', () => {
+      const rowElements = component.find('.row-equal-height');
+      expect(rowElements.length).toBe(0);
+    });
+    it('should add a dl-horizontal and no justified layout classes to the dl element', () => {
+      expect(component.find('.dl-horizontal').length).toBe(1);
+      expect(component.find('.text-xs-left').length).toBe(0);
+      expect(component.find('.text-sm-right').length).toBe(0);
+    });
+    it('does not have dl element in bold', () => {
```
andreapiras: As discussed this should be decided by the data passed in not by the component

src/definitionList/DefinitionList.spec.js
```diff
@@ -0,0 +1,157 @@
...
+      expect(rowElements.length).toBe(0);
+    });
+    it('should add all justified layout classes to the dl element', () => {
+      expect(component.find('.dl-horizontal').length).toBe(1);
+      expect(component.find('.text-xs-left').length).toBe(1);
+      expect(component.find('.text-sm-right').length).toBe(1);
+    });
+
+    it('does not have dl element in bold', () => {
+      expect(component.find('.font-weight-bold').length).toBe(0);
```
andreapiras: As discussed this should be decided by the data passed in not by the component

src/definitionList/DefinitionList.spec.js
OFFLlNE: For Vertical layout and !narrow this is the case, yes.

src/definitionList/DefinitionList.spec.js
OFFLlNE: Yup, will remove the test from here and adjust it to be nice and sexy in the child ;)

src/definitionList/DefinitionList.js
```diff
@@ -59,7 +61,14 @@ class DefinitionList extends PureComponent {
                 {!field.group && (
                   <dl className={this.getClassNameForLine()}>
                     <dt>{field.title}</dt>
-                    <dd className="text-word-break">
+                    <dd
+                      className={classNames(
+                        field.bold && layout === LAYOUTS[2]
```
andreapiras: As discussed this should be decided by the data passed in not by the component

PR #135, opened 2019-04-23
Fix select component animation

Tankenstein:
This adds support for animations in `Select` again by not unmounting the dropdown right away. We could've just kept it hidden, but i also used a `Transition` to unmount it after the animation is done.
Also refactors some things in docs and building classnames in Select.


docs/Docs.js
```diff
@@ -25,7 +25,7 @@ import DefinitionListDocs from './DefinitionListDocs';
 import UploadDocs from './UploadDocs';
 
 const Docs = () => (
-  <Fragment>
+  <StrictMode>
```
andreapiras: image
We need to either fix those errors or open a ticket to fix them :)

src/select/Select.js
```diff
@@ -23,6 +25,10 @@ function stopPropagation(event) {
   // document listener does not use SyntheticEvents
 }
 
+const BOOTSTRAP_DROPDOWN_ANIMATION_TIME = 200;
+
+const defer = fn => setTimeout(fn, 0);
+
 export default class Select extends Component {
```
andreapiras: This can probably be change too to PureComponent

src/select/Select.js
```diff
@@ -345,64 +351,93 @@ export default class Select extends Component {
...
-
-    const dropdownMenu = this.style('dropdown-menu');
-    const dropdownRight = this.props.dropdownRight
-      ? this.style(`dropdown-menu-${this.props.dropdownRight}-right`)
-      : '';
-    const dropdownWidth = this.props.dropdownWidth
-      ? this.style(`dropdown-menu-${this.props.dropdownWidth}`)
-      : '';
-    const dropdownClass = `${dropdownMenu} ${dropdownRight} ${dropdownWidth}`;
+    const s = this.style;
```
andreapiras: Personal preference but I would put a more descriptive name here

src/select/Select.js
```diff
@@ -345,64 +351,93 @@ export default class Select extends Component {
...
-            {!required && !canSearch ? this.renderPlaceHolderOption() : ''}
-            {canSearch ? this.renderSearchBox() : ''}
-            {this.renderOptions()}
-          </ul>
+        {animationState => (
+          <div // eslint-disable-line jsx-a11y/no-static-element-interactions
+            className={groupClass}
+            ref={this.dropdownMenuRef}
+            onKeyDown={this.handleKeyDown}
+            onTouchMove={this.handleTouchStart}
```
andreapiras: should this be handleTouchMove?

src/select/Select.js
```diff
@@ -345,64 +351,93 @@ export default class Select extends Component {
...
+            <button
+              disabled={disabled}
+              className={buttonClass}
+              type="button"
+              id={id}
+              aria-expanded={open}
+              onClick={this.handleButtonClick}
+            >
+              {this.renderButtonInternals()}
+              <span className={this.style('caret')} />
```
andreapiras: s?

docs/Docs.js
Tankenstein: Doesn't block this PR though :)

src/select/Select.js
Tankenstein: No it can't, referential equality will not work for `options`

src/select/Select.js
Tankenstein: This was for shortening all the stuff below 😅 am fine to change it back

docs/Header.js
```diff
@@ -1,53 +1,21 @@
...
+    <section className="section row">
+      <div className="col-md-6">
+        <h1>
+          Components{' '}
+          <span role="img" aria-label="none">
+            🎉
+          </span>{' '}
+          <small>v{npmPackage.version}</small>
+        </h1>
+        <p>TransferWise design system components in react.</p>
```
dozortsev: should we mention that's called Neptune?)

docs/Docs.js
andreapiras: Correct but would be better to have them addressed in a ticket :)

Tankenstein: Good idea 👍  https://transferwise.atlassian.net/jira/software/projects/DS/boards/277?selectedIssue=NP-49

PR #136, opened 2019-04-25
Doc per folder

andreapiras:


docs/utils/loadDocs/index.js
```diff
@@ -0,0 +1,4 @@
+// In this two lines we go through the src folder and load all the files with extension .docs.js into an array that we
```
Tankenstein: Don't think the comment is necessary, if you want to document it, i would make two named functions instaed

src/dateInput/DateInput.docs.js
```diff
@@ -1,8 +1,8 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 
-import { Checkbox, Select, DateInput } from '../src';
+import { Checkbox, Select, DateInput } from '..';
 
-export default class DateInputDocs extends Component {
+export default class DateInputDocs extends PureComponent {
```
Tankenstein: Why the PureComponents everywhere? It's not a straight-up performance win. Unless you are actively having perf trouble/optimising, no reason to use them: https://twitter.com/dan_abramov/status/759383530120110080?lang=en

docs/utils/loadDocs/index.js
andreapiras: Ahhhhhh George asked me to add a comment about the function :sweat_smile:

PR #139, opened 2019-04-29
Remove touchstart from Select and Popover

KarlEST:
Please play around in browserstack to make sure that nothing(`Select, Popover, MoneyInput`) is broken(I did it myself as well but better to have more eyes on it): https://transferwise.github.io/components/branch/remove-touchstart/index.html

*Problem:* https://github.com/transferwise/components/pull/132 We tried to solve problem where people couldn't close `Select` or `Popover` on ios web. For that we introduced `touchstart` for ios devices. With that came a new bug where people clicked/touched items in `Select`. `Select` dropdown closed but nothing was selected because the close was called before selecting anything.  e.g. people couldn't select currencies in Calculator(`MoneyInput` uses `Select` internally).

NB! Ios devices recognise clicks only on <button> and <a> otherwise you have `touch` events.

*Solution:* You can add `cursor: pointer` css to elements and that makes them clickable in ios(because logic 😕 ). So we add `cursor: pointer` to `<html>` when ios device opens `Select` or `Popover` and remove it when they close it. 
Why are we adding click listener to `document` and not `body`? Because for some unknown reason `body` listener event fires before React `Synthetic events` and we can't stop the bubbeling/propagation so body `click` event gets fired all the time. This means we have to add class to `html` root element and then listen to `document` clicks and stop propagation if the click is inside `Select`.

TLDR: Instead of `touchstart` on ios we add `cursor: pointer` to `html` root element and listen to `document` clicks.


src/popover/Popover.js
```diff
@@ -49,14 +51,21 @@ export default class Popover extends Component {
 
   open = () => {
     this.setState({ isOpen: true });
-    document.addEventListener(getEventType(), this.closePopoverOnOutsideClick, true);
+
+    if (isIosDevice()) {
+      document.documentElement.classList.add('ios-click');
```
Tankenstein: I would move dom operations to a separate file, you can mock them more easily then

src/common/domHelpers/documentIosClick.js
```diff
@@ -0,0 +1,13 @@
+import { isIosDevice } from '../deviceDetection';
+
+export function addClickToDocumentOnIos() {
```
mikknarusk: not sure about the name. It adds a class, not really a "click"

OFFLlNE: I like this comment

PR #140, opened 2019-04-30
Select.js mobile animation fix

mikknarusk:
In mobile screens, we use `animation`, while desktop uses `transition` to animate in the dropdown for Select. 
In iOS browsers, these two coexisting causes weird and inconsistent behaviour. This removes `transition` from `xs` screens


mikknarusk:
> What about tablets like ipad?

The animation behavior is only used in `xs` screens, so no tablets to worry about

PR #141, opened 2019-05-03
Add type check to deviceDetection

OFFLlNE:


src/stepper/deviceDetection.js
```diff
@@ -1,8 +1,10 @@
 function supportsTouchEvents() {
   return !!(
-    window.ontouchstart !== undefined ||
-    navigator.maxTouchPoints ||
-    (window.DocumentTouch && document instanceof window.DocumentTouch)
+    (typeof window !== 'undefined' && window.ontouchstart !== undefined) ||
+    (typeof navigator !== 'undefined' && navigator.maxTouchPoints) ||
+    (typeof window !== 'undefined' &&
+      window.DocumentTouch &&
+      document instanceof window.DocumentTouch)
```
dozortsev: this logic became quite hard to read and understand,

could you please split _sub-checks_ on variables, this should simplify understand of this _check_:

example:

```javascript

const onTouchStartIsDefined = typeof window !== 'undefined' && window.ontouchstart !== undefined;
const maxTouchPointsIsDefined = typeof navigator !== 'undefined' && navigator.maxTouchPoints;
...

return onTouchStartIsDefined || maxTouchPointsIsDefined || ...

```

OFFLlNE: Made changes ;) Take a look and lmk if it could be done even better. I could create a function or a const for window typeof check but not sure if that is a good idea

Did not make changes in `userAgentSuggestsTouchDevice` because it does not really make it much better :D

src/stepper/deviceDetection.js
```diff
@@ -1,9 +1,12 @@
...
-    (window.DocumentTouch && document instanceof window.DocumentTouch)
-  );
+  const ontouchstartIsDefined = typeof window !== 'undefined' && window.ontouchstart !== undefined;
+  const maxTouchPointsIsDefined = typeof navigator !== 'undefined' && navigator.maxTouchPoints;
+  const documentTouchIsDefined =
+    typeof window !== 'undefined' &&
+    window.DocumentTouch &&
+    document instanceof window.DocumentTouch;
+
+  return !!(ontouchstartIsDefined || maxTouchPointsIsDefined || documentTouchIsDefined);
```
dozortsev: why `!!` is needed here?

src/stepper/deviceDetection.js
```diff
@@ -1,9 +1,12 @@
 function supportsTouchEvents() {
-  return !!(
-    window.ontouchstart !== undefined ||
-    navigator.maxTouchPoints ||
-    (window.DocumentTouch && document instanceof window.DocumentTouch)
-  );
+  const ontouchstartIsDefined = typeof window !== 'undefined' && window.ontouchstart !== undefined;
```
dozortsev: please use CamelCase style for this variable

src/stepper/deviceDetection.js
OFFLlNE: Probably just in case to make this into a boolean

src/stepper/deviceDetection.js
OFFLlNE: I was thinking, but is it correct to do that since it is actually named with all lower case?

src/stepper/deviceDetection.js
dozortsev: ah, indeed, incase of `maxTouchPointsIsDefined ` it can be assigned by value of `navigator.maxTouchPoints`

src/stepper/deviceDetection.js
dozortsev: imho it looks weird, I would keep it consistent with other two variable and use CC style

but anyway it's minor thing... up to you

src/stepper/deviceDetection.js
OFFLlNE: + it was there before :D

src/stepper/deviceDetection.js
OFFLlNE: Totally agree that it looks weird :D

PR #142, opened 2019-05-03
Fix <RadioButton> click handling

domhorvathtw:
### What?
This PR fixes the issue of `<RadioButton/>` not supporting clicks on the circular `<button>` element.

**Before**
radiobutton1

**After** 
radiobutton2

### Extra
Updated the docs for **Radio** as well. Radio buttons are never alone, they should represent groups of options where the user can only select one. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio


src/radio/Radio.docs.js
```diff
@@ -20,62 +20,90 @@ export default class RadioDocs extends Component {
...
-              <div className="m-t-3" />
-              <Checkbox
-                label="Disabled?"
-                onChange={this.createStateLink('disabled')}
-                checked={this.state.disabled}
-              />
+              <div className="row">
+                {[1, 2, 3].map(index => (
+                  <Fragment key={index}>
+                    <div className="col-md-12" style={{ marginBottom: '16px' }}>
```
dozortsev: IHMO we should have _RadioGroup_ component which will does rendering of >1 _Radio_'s and this CSS style, because only one _Radio_ is useless..

src/radio/Radio.docs.js
```diff
@@ -1,12 +1,12 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import { Radio, Checkbox } from '..';
 
 export default class RadioDocs extends Component {
   constructor(props) {
```
andreapiras: constructor is redundant :)

src/radio/Radio.docs.js
domhorvathtw: Agree, but that should be another component. Inspiration: https://blueprintjs.com/docs/#core/components/radio.

This PR is focusing on fixing an existing bug for apps that are using the <Radio> component

dozortsev: not sure this should be part of this PR though

src/radio/Radio.docs.js
domhorvathtw: why is it redundant? It's implemented to initialize state

PR #143, opened 2019-05-03
Add Button component

domhorvathtw:
## Initial Button component

image
image
image
 
---

### Following design specs
https://transferwise.atlassian.net/wiki/spaces/DS/pages/859604203/Buttons

### Variants
- Primary
- Send
- Default
- Error

### Sates
- Default
- Loading 
- Disabled

### Different sizes 
- Extra small 
- Small 
- Medium (default) 
- Large

### Block option
Added isBlock flag

### Colours
Colour variables should get updated in bootstrap variables
https://github.com/transferwise/bootstrap/blob/master/less/variables.less

**Get new brand colour codes here:** 
https://app.goabstract.com/share/38662c17-b93d-454a-8a45-ea53b92e58f6

### Button layout variants
Circle, Dropdown and Split Button variants will be implemented in separate PR.


src/button/Button.docs.js
```diff
@@ -0,0 +1,138 @@
...
+export default class ButtonDocs extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      label: 'Send money',
+      stateSelection: { value: 'default', label: 'default' },
+      variant: 'pay',
+      size: 'md',
+      block: false,
+      onClick: () => null,
```
dozortsev: I would add some action here e.g `alert(..)` (we use it in a few other component's docs)

src/button/Button.docs.js
```diff
@@ -0,0 +1,138 @@
...
+                  <label htmlFor="button-state-selector" className="control-label">
+                    State
+                  </label>
+                  <Select
+                    id="button-state-selector"
+                    required
+                    selected={stateSelection}
+                    options={STATES.map(state => ({
+                      value: state,
+                      label: !state ? 'default' : state,
```
dozortsev: please avoid using `!`, plus this statement could be simplified `state && 'default'`

src/button/Button.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+export const BUTTON_VARIANTS = ['primary', 'pay', 'secondary', 'danger'];
+export const BUTTON_SIZES = ['xs', 'sm', 'md', 'lg'];
+export const BUTTON_STATES = ['default', 'loading', 'disabled'];
```
dozortsev: why we need expose this constants?

src/button/Button.docs.js
```diff
@@ -0,0 +1,138 @@
...
+    };
+  }
+
+  createStateLink(name) {
+    return value => this.setState({ [name]: value });
+  }
+
+  render() {
+    const { label, stateSelection, size, variant, block, onClick } = this.state;
+    const buttonLabel = label === '' ? 'Send Money' : label;
```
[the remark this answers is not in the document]
domhorvathtw: fixed, thanks!

src/button/Button.docs.js
domhorvathtw: Added alert

src/button/Button.js
domhorvathtw: good catch, they were used in documentation as well

domhorvathtw:
The `<button>` element supports pressing enter when it has the focus. Select it with the tab and press enter and you'll see it works. The submit example is not necessary in my opinion, that's a specific use case, that you can achieve by using the button with the provided props (variants and states).

src/button/Button.docs.js
domhorvathtw: that's actually completely unecessary, good catch, thanks!

StevePole:
Could you make it so the text doesn't move when in loading state?

image

StevePole:
Could you make it so the text doesn't move when loading?

button-loading

domhorvathtw:
> Could you make it so the text doesn't move when loading?
> 
> button-loading

Good catch, would have been annoying to have the text jumping back and forth between state changes. Since I added custom CSS, I also fixed the loader position and sizing for `xs` buttons.

src/button/index.js
```diff
@@ -0,0 +1,3 @@
+import Button from './Button';
```
oliverviljamaa: How about `export { default } from './Button';`?

src/button/Button.js
```diff
@@ -0,0 +1,52 @@
...
+    >
+      {label}
+      {isLoading && <span className={`btn-loader ${block ? '' : 'm-l-2'}`} />}
+    </button>
+  );
+};
+
+Button.propTypes = {
+  variant: Types.oneOf(BUTTON_VARIANTS),
+  size: Types.oneOf(BUTTON_SIZES),
```
oliverviljamaa: How about exposing the `Variant`, `Size`, and `State` as enums for consumers?

So I could do:
```js
<Button variant={ButtonVariant.PAY} size={ButtonSize.SMALL} buttonState={ButtonState.DISABLED}>
```

oliverviljamaa: We could use the same enum-like objects to avoid repetition and more fragile string checks in this file as well 😊

src/button/Button.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <button
+      type="button"
+      className={classes}
+      onClick={e => onClick(e)}
+      disabled={isDisabled || buttonState === 'loading'}
+    >
+      {label}
+      {isLoading && <span className={`btn-loader ${block ? '' : 'm-l-2'}`} />}
```
oliverviljamaa: What would you think about using `classNames` here?
```suggestion
      {isLoading && <span className={classNames('btn-loader', { 'm-l-2': block })} />}
```

src/button/Button.js
domhorvathtw: Agree 👍

src/button/Button.js
domhorvathtw: fixed

src/button/index.js
domhorvathtw: fixed 🙇

src/button/Button.js
domhorvathtw: Agree 👍  exposed the enum-like objects `ButtonVariants`, `ButtonStates`, `ButtonSizes`

src/button/Button.docs.js
```diff
@@ -0,0 +1,142 @@
...
+    };
+  }
+
+  createStateLink(name) {
+    return value => this.setState({ [name]: value });
+  }
+
+  render() {
+    const { label, stateSelection, size, variant, block, onClick } = this.state;
+    const buttonLabel = label || 'Send Money';
```
oliverviljamaa: Do we want to have a default in the demo? I found that a bit unintuitive, but it may just be me 😀

src/button/Button.docs.js
```diff
@@ -0,0 +1,142 @@
+import React, { Component } from 'react';
+import { Button, Select, Checkbox } from '..';
+import { ButtonSizes, ButtonStates, ButtonVariants } from './Button';
+
+export default class ButtonDocs extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
oliverviljamaa: Could we define `state` as a class field declaration and avoid the constructor here?

src/button/Button.docs.js
```diff
@@ -0,0 +1,142 @@
...
+  variant={"${variant}"}
+  state={"${stateSelection.value}"}
+  block={${block}}
+  onClick={this.handleClick}
+/>`}
+              </pre>
+            </div>
+            <div className="col-md-6">
+              <div className="row">
+                <div className="col-md-12">
```
oliverviljamaa: If Bootstrap grid is 12 by default and this class makes it 12 from `md` onwards, it seems that it's just always 12 and we can remove the `div.row` and the `div.col-md-12`. But it's possible I'm just looking at it wrong.

src/button/Button.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+import './Button.less';
+
+export const ButtonVariants = {
```
oliverviljamaa: Thanks for adding these!
As they're enum-like structures and enums are usually named in singular, what do you think of having them in singular here as well?

https://stackoverflow.com/questions/15755955/naming-of-enums-in-java-singular-or-plural

src/button/Button.js
```diff
@@ -0,0 +1,68 @@
...
+  Large: 'lg',
+};
+
+export const ButtonStates = {
+  Default: 'default',
+  Loading: 'loading',
+  Disabled: 'disabled',
+};
+
+const Button = ({ label, buttonState, size, variant, block, onClick }) => {
```
oliverviljamaa: I may have missed some conversations about it, but is there a reason why we've gone with `variant` here rather than `type`, for example?

I feel _variant_ is something that _differs_ from some norm, but for us, every button type is equal, even the default button has a type.

src/button/Button.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    wrapper.setProps({ block: true });
+    expect(wrapper.find('button').hasClass('btn-block')).toBe(true);
+  });
+
+  it('calls click handler on button click if not disabled', () => {
+    expect(props.onClick).not.toBeCalled();
+    wrapper.simulate('click');
+    expect(props.onClick).toBeCalled();
+  });
+});
```
oliverviljamaa: Great tests. 👌

src/button/Button.docs.js
domhorvathtw: do you mean not having a default label like "Send Money"?
if we don't have it, how would the button look like initially?

src/button/Button.docs.js
domhorvathtw: I don't have a strong preference about state initialisation? I quickly checked, we use both constructor and class property to initialise state in components and component's docs. I feel like it's another discussion about creating / following a company wide react code style, what do you think?

src/button/Button.docs.js
domhorvathtw: nope, you're right, my bad. Fixed it!

src/button/Button.js
domhorvathtw: Sure thing, makes sense

src/button/Button.spec.js
domhorvathtw: 👍

src/button/Button.js
domhorvathtw: Not really, I don't want to stick to `variant`. Let's use `type` and `variant` could be used when a button is different in shape or functionality, e.g. Circular, Dropdown, Split buttons. What do you think?

src/button/Button.docs.js
oliverviljamaa: You're right, it's something that should be enforced by ESLint 🙂 I normally just prefer the shorter version, so class field declarations, but there's really no big difference, so let's leave it as is here.

src/button/Button.docs.js
oliverviljamaa: It's already in the state, so by default, it would say `Send money`. However, if we remove it from here (`render`), then the button will be empty when I erase the word from the docs input, which is expected, imo.

src/button/Button.js
oliverviljamaa: Agreed, I couldn't even think of the added benefit of being able to use `variant` in future scenarios!

src/button/Button.docs.js
domhorvathtw: True, removed it ✅

PR #144, opened 2019-05-09
Np 51 radio group

andreapiras:
This Component renders a group of radio options. It has been created as part of discussion https://github.com/transferwise/components/pull/142#discussion_r280745599


src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
+import classNames from 'classnames';
+import { RadioGroup } from '..';
+
+const radios = [
```
dozortsev: upper case?

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+const getExtraDocs = () => (
+  <div>
+    <p className=" m-t-5">
+      This component render a simple group of radio options, and broadcast the index of selected
+      option onChange
+    </p>
+  </div>
```
dozortsev: why this piece should be separated? :)

dozortsev: btw, I think things like `onChange` should be wrapped in `<code>..` tag or something like this..

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,63 @@
...
+      {`<RadioGroup 
+ radios={[options]} 
+ onChange={this.handleOnChange}
+/>`}
+    </pre>
+  );
+
+  handleOnChange = value => {
+    // eslint-disable-next-line
+    console.log('Value selected', value);
```
dozortsev: why not show and update `value` in docs?

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,54 @@
...
+        checked: Types.bool,
+        disabled: Types.bool,
+      }),
+    ).isRequired,
+    onChange: Types.func.isRequired,
+  };
+
+  getChecked = () => {
+    const { radios } = this.props;
+    return radios.findIndex(radio => radio.checked);
```
dozortsev: `findIndex` seems [not supported by IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Browser_compatibility)

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    const { checked } = this.state;
+    const { radios } = this.props;
+    return radios.length ? (
+      <Fragment>
+        {radios.map(({ id, label, name, disabled }, index) => (
+          <Radio
+            id={id || `radio-${index}`}
+            // eslint-disable-next-line react/no-array-index-key
+            key={index}
```
dozortsev: does this eslint comment still should be here since you added `key` attr?

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,54 @@
...
+
+  handleOnChange = checked => {
+    const { onChange } = this.props;
+    this.setState({ checked }, onChange && onChange(checked));
+  };
+
+  render() {
+    const { checked } = this.state;
+    const { radios } = this.props;
+    return radios.length ? (
```
dozortsev: I would check this in constructor (or in one of component lifecycle methods) and throw an error if consumer provided less then 2 radios, make sense?

src/radioGroup/RadioGroup.js
andreapiras: This eslint rule prevents to use the index as key, but in this [case](https://github.com/yannickcr/eslint-plugin-react/issues/1123) you can override and I think in this case is fine

src/radioGroup/RadioGroup.js
domhorvathtw: Why restrict to 2 radios? I know we talked about the fact that radio buttons are never alone, but I think this check is unnecessary, what do you think about something like this:

```javascript
render () {
  ...
  return (
    <Fragment> // or just use `<>`
       radios.map(radio => <Radio ... > ... </Radio>)
    </Fragment>
  )
}
```
if it's an empty array, we render nothing.

docs/utils/loadDocs/index.js
```diff
@@ -1,4 +1,4 @@
 // In this two lines we go through the src folder and load all the files with extension .docs.js into an array that we
 // can loop after to load the components.
-const req = require.context('../../../src', true, /\.docs\.js$/);
+const req = require.context('../../../src', true, /RadioGroup.docs\.js$/);
```
dozortsev: btw, I think now we don't need this doc https://github.com/transferwise/components/blob/master/src/radio/Radio.docs.js

src/radioGroup/RadioGroup.docs.js
andreapiras: > why this piece should be separated? :)

I'm trying to follow the same pattern as this [](https://github.com/transferwise/components/blob/master/src/upload/Upload.docs.js#L151)
Sometimes this extra doc it s normally the "custom bit". At some point I'd like to centralize those utilities

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v4.21.0
+## Add Radio Group component
+Added radio group to render a 2 or more radio buttons
```
domhorvathtw: **Suggestion:**
Added `<RadioGroup/>` component for rendering multiple (2 or more) radio buttons.

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,110 @@
...
+import classNames from 'classnames';
+import { RadioGroup } from '..';
+
+const RADIOS = [
+  { label: 'Radio1', name: 'name', disabled: false },
+  { label: 'Radio2', name: 'name', disabled: false },
+  { label: 'Radio3', name: 'name', disabled: false },
+];
+
+const KNOBS = {
```
domhorvathtw: Nice 👍

src/radioGroup/RadioGroup.spec.js
```diff
@@ -0,0 +1,103 @@
...
+
+const props = { onChange: jest.fn() };
+
+const RADIOS = [
+  { label: 'Radio1', name: 'name', disabled: true },
+  { label: 'Radio2', name: 'name', disabled: false },
+  { id: 'id-test', label: 'Radio3', name: 'name', disabled: false },
+];
+console.error = jest.fn();
+describe('RadioGroup', () => {
```
domhorvathtw: nice tests 🥇

src/radioGroup/RadioGroup.docs.js
andreapiras: Because it's not a prop it's a broadcasted value and we are currently displaying it as console.log. We can decide to display differently and create a ticket to implement it

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,110 @@
...
+            );
+            return acc;
+          }, [])
+          .join('\n\r  '),
+      )
+      .join();
+
+    return (
+      <pre className="tw-docs-code ">
+        {`<Upload
```
domhorvathtw:
```javascript
<RadioGroup
  radios={[...]}
  onChange={...}
/>
```

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+        name: Types.string.isRequired,
+        disabled: Types.bool,
+      }),
+    ).isRequired,
+    onChange: Types.func.isRequired,
+    checkedRadio: Types.number,
+  };
+
+  static defaultProps = {
+    checkedRadio: 0,
```
domhorvathtw: What should be the default state for a radio button group?
I feel like default state should be `null` or something that indicates that nothing is selected yet and it's up to the user to make the first selection, unless the consumer app decides to set an option preselected for the user.

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,110 @@
...
+              {getExtraDocs()}
+              {this.getDocs()}
+            </div>
+            <div className="col-md-6">
+              <div
+                className={classNames({
+                  ' m-b-4': true,
+                })}
+              >
+                <RadioGroup
```
domhorvathtw: I would add a preview of the selected radio button below these components. So when you play with it an click around, you could see which on is selected.

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+    radios: Types.arrayOf(
+      Types.shape({
+        id: Types.string,
+        label: Types.string.isRequired,
+        name: Types.string.isRequired,
+        disabled: Types.bool,
+      }),
+    ).isRequired,
+    onChange: Types.func.isRequired,
+    checkedRadio: Types.number,
```
dozortsev: hmm, I think instead of having number of checked radio option we better have a value of checked one.. I think in this case component will be more useful

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+    const { onChange } = this.props;
+    this.setState({ checked }, onChange && onChange(checked));
+  };
+
+  render() {
+    const { radios } = this.props;
+
+    if (!radios || radios.length < 2) {
+      // eslint-disable-next-line
+      console.error('At least two radio options should be provided');
```
dozortsev: why not throwing an error instead?

src/radioGroup/RadioGroup.js
domhorvathtw: I think the confusion is coming from the naming. If we named it `defaultSelection` or something. Then you would not expect the prop to update, because it's only for the first render, and you decide what you want to do. Every user interaction will be reflected in the onChange prop and it handler callback.

src/radioGroup/RadioGroup.js
andreapiras: Because I think it's better if we don't break the whole page if that components errors

src/radioGroup/RadioGroup.js
andreapiras: checkedRadio is the value of the selected one :)

andreapiras: I just updated it 😂

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+    radios: Types.arrayOf(
+      Types.shape({
+        id: Types.string,
+        label: Types.string.isRequired,
+        name: Types.string.isRequired,
+        disabled: Types.bool,
+      }),
+    ).isRequired,
+    onChange: Types.func.isRequired,
+    selectedValue: Types.number,
```
dozortsev: why `selectedValue` can be only a `number` type? I think it could be as well as `string`, as `shape` and so on...

use case: radio-group with two options "Male" and `Female`, once user select one of them `selectedValue` should be equal one them respectively

does this make sense?

dozortsev: example with native HTML: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_radio

dozortsev: shit, I probably see why you didn't implement it in this way... [it's because the `<Radio />` component doesn't have any `value` prop...](https://github.com/transferwise/components/blob/master/src/radio/Radio.js#L16-L23) which I think is bad

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+
+    if (!radios || radios.length < 2) {
+      // eslint-disable-next-line
+      console.error('At least two radio options should be provided');
+    }
+
+    const { checked } = this.state;
+    return radios ? (
+      <Fragment>
+        {radios.length > 1 &&
```
dozortsev: this two checks could be combined, e.g:

```javascript
return radios && radios.length > 1 && (
  <Fragment>
  ...
```

src/radioGroup/RadioGroup.js
domhorvathtw: > why `selectedValue` can be only a `number` type? I think it could be as well as `string`, as `shape` and so on...
> 
> use case: radio-group with two options "Male" and `Female`, once user select one of them `selectedValue` should be equal one them respectively
> 
> does this make sense?

I would disagree with this. When you would use the radio options with your example (options: `['Male', 'Female]`) and the user selected something, the `onChange` handler is the way to update your **container component's** logic. The **<RadioGroup/>** component's internal state should handle the selectedValue, not the container component that uses it. 

IMO, the `selectedValue` prop is used for initializing a selection for the user if needed, therefore it should be optional.

dozortsev: ## Update

we had a call and agreed to change _<Radio />_ component so that it has `value` prop then then pass this value in [`handleOnChange` function](https://github.com/transferwise/components/pull/144/files#diff-dd85e41eab7f2c39e79620740de2df14R25).

src/radioGroup/RadioGroup.js
domhorvathtw: 👍

src/common/RadioButton/RadioButton.js
```diff
@@ -37,13 +35,15 @@ RadioButton.propTypes = {
...
   disabled: Types.bool,
+  value: Types.string,
 };
 
 RadioButton.defaultProps = {
   checked: false,
   onChange: () => {},
   disabled: false,
   id: null,
+  value: '',
```
dozortsev: maybe just `null` as [here](https://github.com/transferwise/components/pull/144/files#diff-084edf5e28d5cb92e93c48930a1c39dfR25) you pass `null` as sort of "default" value

src/radio/Radio.js
```diff
@@ -14,18 +14,22 @@ const Radio = ({ label, id, ...otherProps }) => (
...
-  label: Types.string.isRequired,
+  checked: Types.bool,
+  disabled: Types.bool,
   id: Types.string,
+  label: Types.string.isRequired,
   name: Types.string.isRequired,
   onChange: Types.func.isRequired,
-  checked: Types.bool,
-  disabled: Types.bool,
+  secondary: Types.string,
```
dozortsev: what this need for?

dozortsev: I don't see where you use it in _Radio_ component

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,63 @@
...
+      Types.shape({
+        id: Types.string,
+        value: Types.string,
+        secondary: Types.string,
+        label: Types.string.isRequired,
+        disabled: Types.bool,
+      }),
+    ).isRequired,
+    onChange: Types.func.isRequired,
+    selectedValue: Types.oneOfType([Types.string, Types.number]),
```
dozortsev: types of `selectedValue` and `value` in `radios` should be the same

src/radioGroup/RadioGroup.js
```diff
@@ -0,0 +1,63 @@
...
+
+  static defaultProps = {
+    selectedValue: null,
+  };
+
+  state = { selectedValue: this.props.selectedValue };
+
+  handleOnChange = selectedValue => {
+    const { onChange } = this.props;
+    this.setState({ selectedValue }, onChange && onChange(selectedValue));
```
dozortsev: I guess it's minor thing bug [React Official doc](https://reactjs.org/docs/react-component.html#setstate) says that better use `componentDidUpdate` rather than callback param in  `setState`.

dozortsev: but not a big deal, JFYI

src/radio/Radio.js
andreapiras: There's a ticket for that ,for more info check this https://transferwise.atlassian.net/jira/software/projects/DS/boards/277?selectedIssue=NP-51

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,140 @@
...
+    value: 'radio-3',
+    label: 'Radio3',
+    secondary: 'Secondary line 3',
+    name: 'name',
+    disabled: false,
+  },
+];
+
+const KNOBS = {
+  firstStepKnobs: [
```
dozortsev: this prop is seems redundant here

src/radioGroup/RadioGroup.docs.js
```diff
@@ -0,0 +1,140 @@
...
+      type: 'input',
+      label: 'radios',
+      state: 'radios',
+      defaultState: 'Array of radio objects',
+    },
+    {
+      type: 'input',
+      label: 'name',
+      state: 'name',
+      defaultState: 'radio-group',
```
dozortsev: could you make disabled one of them? to showcase all features/states of this component

src/radio/Radio.js
dozortsev: ok, make sense

src/radioGroup/RadioGroup.docs.js
andreapiras: This is not a prop is just how automated docs works. It's inline with https://github.com/transferwise/components/blob/master/src/upload/Upload.docs.js#L22
but i wouldn't worry too much about it as it may be replaced soon with Carlos docs anyway

PR #146, opened 2019-05-13
MessageControl component

dozortsev:
## tl;dr desc

This component build so way that it could be used for messaging in _Field_ component as well as a simple components as _Input_, _Checkbox_, _RadioOption_ etc (wrapped inside `<div class="form-group has-info | has-error..">..`).

## Demo

https://transferwise.github.io/components/branch/control-message-component-NP-59/index.html#message-control

### Content types

- single message
- list of messages
- image
- element

### Inside Field component

- help text (Visibility: focus, Type of message: info, Content types: all)
- warning (Visibility: always, Type of message: warning, Content types: single (but could be all)) and we do not show help text.
- error (Visibility: always*, Type of message: error, Content types: single, list of messages (but could be all)) if there is an error we do not show warning or help text.

### With simple form controls (e.g Checkbox, input, RadioOption etc.) (non DF)

.. (check demo)

## Jira

https://transferwise.atlassian.net/browse/NP-59


src/common/validation/type-validators/index.js
```diff
@@ -0,0 +1,20 @@
+function isString(value) {
```
dozortsev: [source](https://github.com/transferwise/styleguide-components/blob/master/src/json-schema/validation/type-validators/index.js)

src/index.js
```diff
@@ -21,3 +21,4 @@ export { default as Upload } from './upload';
 export { default as DefinitionList } from './definitionList';
 export { default as Button } from './button';
 export { default as RadioGroup } from './radioGroup';
+export { default as MessageControl } from './messageControl';
```
dozortsev: not sure about name yet, could be just `<Message />` because it's not only for _form control_

matallo: well, we're already using `<Message>` for translations

src/messageControl/MessageControl.js
```diff
@@ -0,0 +1,87 @@
...
+
+export const MessageType = {
+  Info: 'info',
+  Warning: 'warning',
+  Error: 'error',
+};
+
+const MessageControl = ({ type, content, focusVisible, children }) => {
+  function getMessageContent() {
+    if (isString(content)) {
```
StevePole: Not entirely sure about this polymorphic approach.  Doesn't seem too clear what can be passed in, and perhaps we're just being too flexible.  We could for example always ask for an array of strings, It would still trivially support a single string, and maps could be converted to arrays quite simply.

We would need to make provision for images as well.  I think wewould only ever support one image, so that could be a separate prop?

src/messageControl/MessageControl.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  const messageContent = getMessageContent();
+  const groupClasses = classNames('form-group', {
+    'has-info': type === MessageType.Info && !focusVisible,
+    'has-error': type === MessageType.Error,
+    'has-warning': type === MessageType.Warning,
+  });
+  const messageClasses = classNames(
+    'alert',
```
StevePole: I believe we have a separate task for creating an alert component, not sure if anyone started on it, but I wonder if this should be using that.

Something like:
```
<Alert type="warning" dismissable="false" size="sm"> 
  messages.map(message => { message })
</Alert>
```

dozortsev: I'm sure that [ticket](https://transferwise.atlassian.net/jira/software/projects/NP/boards/277?selectedIssue=NP-7) about another alert :)

It's more about something like this:

(https://transferwise.github.io/neptune-css/components/Alerts/)

Screen Shot 2019-05-15 at 9 53 04 AM

and yeah we plan to have a React component for this `<Alert>` just like you showed code above (no one started working on it yet, I plan to do once I finish with this one).

src/messageControl/MessageControl.js
dozortsev: thanks for feedback

> ..We could for example always ask for an array of strings, It would still trivially support a single string..

I'm sort of agree but what the problem with current approach where we identify the type of provided value?

not sure about following:

> ..and maps could be converted to arrays quite simply.

how following could be passed as an array of strings?

```json
{
 "required": "Required",
 "pattern": "..",
 ..
}
```


> Not entirely sure about this polymorphic approach. Doesn't seem too clear what can be passed in, and perhaps we're just being too flexible.

I maybe just need refactor `getMessageContent` method, try to make it shorter and simpler to understand, and strengthen type checks, wdyt?

dozortsev: JFYI [here is how we'll use this component inside _Field_ component](https://github.com/transferwise/components/pull/131/files#diff-08e06f629432fae9b73c060b4e20a2b7R196) so you see how those [requirements](https://transferwise.slack.com/archives/C2XG7R3EZ/p1557392854060100?thread_ts=1557319315.054200&cid=C2XG7R3EZ) will be implemented there.

Please don't code review _Field_ component PR ☝️ it's not ready and could slightly changed/improved/refactored, just showing this for more context and hopefully better understanding.

src/common/validation/type-validators/index.js
andreapiras: those could all be arrow functions

src/messageControl/MessageControl.docs.js
```diff
@@ -0,0 +1,212 @@
...
+];
+
+export default class MessageControlDocs extends Component {
+  static printMessageComponentProps(object) {
+    return Object.keys(object)
+      .map(key => `${key}="${object[key]}"\n`)
+      .join('\t');
+  }
+
+  constructor(props) {
```
andreapiras: constructor is redundant here

src/messageControl/MessageControl.js
StevePole: I'm a bit confused.  It looks like Messages is a parent of the label and control.  I would expect it to be a sibling.  I was expecting the (rendered) DOM to look something more like this:
```
<Field>
  <Label ...  />
  <FormControl 
    onFocus="showHelp()" 
    onBlur="hideHelp()" 
    onChange="validateAndBroadcast(value)">
    <input type="text" ... />
  </FormControl
  <FormMessages type="messageType" messages="messages">
    <Alert type="..." dismissable="false" size="sm"> ... </Alert>
  </FormMessages
</Field>
```
with Field providing the plumbing to make sure the right messages are provided to the message component.

StevePole: > how following could be passed as an array of strings?

Why do you need to pass it?  In the Angular version we're using CSS to hide, so those names are important, but in the React I was anticipating that we'd only pass in the messages that are relevant, so we'd stop using CSS for visibility, and wouldn't need the keys.

Generally, I just think it would be simpler (both in code and usage) to expect the arguments to be of one type.  Reading the code I didn't even realise that you were going to use the map that way.  It shouldn't really matter how the consumer is going to use it, we should be designing this to provide the best interface for what it is.  If we get that right, then it should be fine for consumers as well.

src/messageControl/MessageControl.js
StevePole: It is about that, but isn't that the same?  Yes, this one has a bit more padding (size), and it has the option to be dismissible, and it has the option to have a pointer.  Our case is the same, we just always choose `size="sm" dismissable="false" pointer="false"`

src/messageControl/MessageControl.spec.js
```diff
@@ -0,0 +1,141 @@
...
+          Lorem ipsum <b className="test">dolor</b> sit.
+        </Fragment>
+      ),
+    });
+
+    const alert = component.find('.alert');
+    expect(alert.find('b')).toHaveLength(1);
+  });
+
+  it('wont render alert if content is empty', () => {
```
andreapiras: won't

src/messageControl/MessageControl.js
dozortsev: re. passing an object

I agreed, it's indeed make sense pass only array of strings in such case, I addressed this change, refactored `getMessageContent` now it won't handle object value.

re.

> ..I'm a bit confused. It looks like Messages is a parent of the label and control. I would expect it to be a sibling. ..

It wraps _children_ component to be able to manage and add necessary CSS classes like `.form-group` (without it, alert will have padding) plus `.has-error`, `has-warning`.. which change border color of children component (_Field_, _Checkbox_ or _Input_ etc.). Does this make sense?

JFYI similarly it's [implemented on Angular](https://github.com/transferwise/styleguide-components/blob/master/src/forms/field/field.html#L1-L4).

re. `<Alert>`

I agree with you to have it as a separate component and use it use it inside _MessageControl_ component, I'm starting work on it.

StevePole: On the management of has-error:

For me it's up to Field to decide if the new value has an error.  If so, it knows what type of error because it gets a list of the failed validations.  It can apply 'has-error' to the form-group, it then has translation strings for different validation types, and it can pass the relevant strings to the message component.

I feel like the alternative is giving the responsibilities of Field (e.g. validation) to the message component, making field redundant.  In my mind the messaging component is a pretty 'dumb' view, rendering what it's been instructed to render, it shouldn't know stuff like how to validate.

dozortsev: ok, refactored it, so now it won't be a parent for control component (_Field_, _Checkbox_).

src/common/validation/type-validators/index.js
```diff
@@ -0,0 +1,13 @@
+const isString = value => value && typeof value === 'string';
```
StevePole: Are we going to import the tests for this?

src/messageControl/MessageControl.js
```diff
@@ -0,0 +1,49 @@
...
+import Types from 'prop-types';
+
+import Alert, { AlertType, AlertSize, AlertArrowPosition } from '../alert/Alert';
+
+import { isString, isArray } from '../common/validation/type-validators';
+
+const MessageControl = ({ type, content }) => {
+  function getMessageContent() {
+    if (isString(content)) {
+      if (isImagePath(content)) {
```
StevePole: What do you think about my suggestion of a separate 'imageUrl' prop?  This approach feels a bit hacky/brittle

src/messageControl/MessageControl.spec.js
```diff
@@ -0,0 +1,132 @@
...
+    console.error = jest.fn();
+    component = shallow(<MessageControl {...props} />);
+  });
+
+  afterEach(() => {
+    console.error.mockClear();
+    jest.clearAllMocks();
+  });
+
+  it('recognized provided content as a simple string and just forward it to the Alert', () => {
```
StevePole: I think the tests would be clearer if you followed the practice of describing the scenario before stating expectations, but it's not a blocker.  I always follow 'given...', 'when...', 'it should...':
```
describe('given a component for rendering validation messages'), () => {
  ...
  describe('when given a simple string', () => {
    let alert, alertProps;
    beforeEach(() => {
      component.setProps({ ... });
      alert = component.find(Alert).at(0);
      alertProps = alert.props();
    });
    it('should render the messages in an alert', () => {
      expect(alert).toBeTruthy();
    });
    it('should correctly configure the alert settings', () => {
      expect(alertProps.size).toEqual(AlertSize.Small);
      expect(alertProps.type).toEqual(AlertType.Info);
      expect(alertProps.arrow).toEqual(AlertArrowPosition.UpLeft);
      expect(alertProps.dismissible).toEqual(false);
    });
    it('should render the string as a message', () => {
      expect(alertProps.children).toBe(props.content);
    });
    it('should no log any errors', () => {
      expect(console.error).not.toHaveBeenCalled();
    });
  });
});
```

src/messageControl/MessageControl.js
dozortsev: IHMO having additional prop for "specific" type of content like image is actually hacky, the interface of component will feels strange to me and confusing, you either have to use `content` prop or `imageUrl` one.. 

I would go two ways, either we have one prop `content` which is generic and accept diff types of content and inside component we have logic which determine how to represent it (current implementation)

or have a prop for each of types of content, `singleLine`, `listOfStrings` (ok, 👈this two could be merged in one `lines`), `imageUrl`, `domContent` ..(maybe something new in future):

```html
<MessageControl
  lines={'foo' | ['bar', 'bar']}
  imageUrl={'image.png'}
  domContent={<Fragment>Lorem <b>lorem</b></Fragment>}
  ..
/>
```

StevePole: I find it hacky because we're trying to interpret the intent of the consumer in an unsafe way.  What if my string was "The file you uploaded, 'selfie.jpg', was too large."  We're going to attempt to render that as an image...

Another issue is that you couldn't show an image and explanatory text at the same time.  Putting data types in the prop interface is weird, but if we just asked for one consistent type we wouldn't need to.  Given I'm not sure we need to pass fragments, an interface like this could work:
```
<MessageControl
  messages={['foo']}
  image="{bar.jpg}" />
```

dozortsev: DM

PR #148, opened 2019-05-15
Alert component

dozortsev:
## Demo

https://transferwise.github.io/components/branch/alert-NP-7/index.html#alert

## Jira

https://transferwise.atlassian.net/browse/NP-7


src/alert/Alert.docs.js
```diff
@@ -0,0 +1,152 @@
...
+            <div className="col-md-6">
+              {/* eslint-disable react/jsx-indent */}
+              <pre className="tw-docs-code">
+                {`<Alert
+    type="${type}"
+    small="${small}"
+    pointer="${pointer}"
+    dismissible="${dismissible}">
+    ${content}
+</Alert>`}
```
andreapiras: Tnhis should be ```/>```

src/alert/Alert.js
```diff
@@ -0,0 +1,55 @@
...
+    'alert-danger': type === 'error',
+    'alert-focus': focusVisible,
+    arrow: pointer,
+    small,
+    'p-x-2': small,
+    'p-y-1': small,
+  };
+
+  return (
+    <div role="alert" className={classNames('alert', 'alert-detach', alertClass)}>
```
andreapiras: Why is 'alert-detach' here? it doesn't appear in the default boostrap component

dozortsev: w/o this CSS class it will look as _attached_ one

Screen Shot 2019-05-15 at 4 53 54 PM

dozortsev: you can reproduce this on demo page (just remove this class from DOM)

src/alert/Alert.js
```diff
@@ -0,0 +1,55 @@
...
+      {dismissible && (
+        <button type="button" className="close" data-dismiss="alert" aria-label="Close">
```
andreapiras: I guess this button should have at least an onClick to actually dismiss the alert

andreapiras:
Documentation looks odd 
image
The knobs should all be on the right just underneath the component

src/alert/Alert.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const Alert = ({ children, type, dismissible, small, pointer, focusVisible }) => {
+  const alertClass = {
+    'alert-success': type === 'success',
+    'alert-info': type === 'info',
+    'alert-warning': type === 'warning',
+    'alert-danger': type === 'error',
+    'alert-focus': focusVisible,
+    arrow: pointer,
```
andreapiras: Not major : I think I'd keep this name as arrow

src/alert/Alert.js
dozortsev: hmm, I'm not sure how it should..

probably we should don't add this class if component is not _small_

StevePole: I don't think it should be based on the size.  Probably we should simply remove the 'attached' styling from Bootstrap and make the detached styling the default, as that was the conclusion from the discussions.  I would be fine if you left out alert-detached for now.  As soon as we make the change in bootstrap they will update.

src/alert/Alert.docs.js
```diff
@@ -0,0 +1,163 @@
...
+      </Fragment>
+    ),
+  },
+];
+
+export default class AlertDocs extends Component {
+  state = {
+    type: AlertType.Warning,
+    arrow: true,
+    small: true,
```
StevePole: Typically most components have a size parameter with sm/md/lg.  I think we should stay consistent with that, even if we do not have a large variant right now.

src/alert/Alert.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export default class Alert extends Component {
+  static propTypes = {
+    type: Types.string,
+    children: Types.node.isRequired,
+    small: Types.bool,
+    dismissible: Types.bool,
+    arrow: Types.bool,
+    focusVisible: Types.bool,
```
StevePole: This is mixing concerns as alerts are not focusable.  I know in this case we want to show/hide based on focus of the input, but that is not the concern of this component. I think we should let the parent figure out whether to show the alert (i.e. not using CSS this time).

src/alert/Alert.js
```diff
@@ -0,0 +1,75 @@
...
+  Success: 'success',
+};
+
+export default class Alert extends Component {
+  static propTypes = {
+    type: Types.string,
+    children: Types.node.isRequired,
+    small: Types.bool,
+    dismissible: Types.bool,
+    arrow: Types.bool,
```
StevePole: Not sure if we should allow position information here, as we do have styles for arrows to appear in a few different spots.

dozortsev: > ..Not sure if we should allow position information here

do you mean position of arrow?

> ..as we do have styles for arrows to appear in a few different spots.

could you please show some examples?

src/alert/Alert.docs.js
dozortsev: ok,

but I think we actually have large one

Screen Shot 2019-05-16 at 11 57 54 AM

we don't have medium right now, agree?

src/alert/Alert.js
dozortsev: > Not sure if we should allow position information here

@StevePole why?

dozortsev: we could refactor this prop so that it will accept position of arrow, this prop will have two functions: 1. tell Alert to show arrow and 2. where show it

wdyt?

StevePole: I think presence of a position is enough to indicate there should an arrow, otherwise no arrow.  So should be possible with one prop.

src/alert/Alert.docs.js
StevePole: Good question 🤔  I would maybe ask in #design-system.  I kind of think this should be the 'default' which would mean mid.  But it really comes down to whether it's more likely that we add a bigger one, or one in between small and this.

dozortsev: https://transferwise.slack.com/archives/C2XG7R3EZ/p1558002932036900

src/alert/Alert.js
```diff
@@ -0,0 +1,111 @@
...
+          'p-y-1': size === AlertSize.Small,
+        };
+
+    function arrowClasses() {
+      if (arrow) {
+        const classes = ['arrow'];
+
+        switch (arrow) {
+          case AlertArrowPosition.BottomCenter:
+            return classes.concat('arrow-bottom', 'arrow-center');
```
andreapiras: extra comma?

src/alert/Alert.js
```diff
@@ -0,0 +1,111 @@
...
+
+export const AlertSize = {
+  Small: 'sm',
+  Large: 'lg',
+};
+
+export const AlertArrowPosition = {
+  UpLeft: 'up-left',
+  UpRight: 'up-right',
+  UpCenter: 'up-center',
```
StevePole: The opposite of 'up' is 'down', and the opposite of 'bottom' is 'top', let's be consistent on one pair.

PR #149, opened 2019-05-16
Moved generateDocs into common folder

andreapiras:


docs/utils/generateDocs/generateDocs.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+const getDocs = (componentName, knobs, componentThis) => {
```
domhorvathtw: is `getDocs` used for generating the `JSX` code block? I find the **getDocs** naming a bit confusing. Can we think about some alternatives? 

Maybe: `generateCodeBlock`

src/common/RadioButton/RadioButton.js
```diff
@@ -43,7 +43,7 @@ RadioButton.defaultProps = {
   onChange: () => {},
   disabled: false,
   id: null,
-  value: null,
```
domhorvathtw: this doesn't seem to be related to this PR? Or is it?

src/radio/Radio.js
```diff
@@ -29,7 +29,7 @@ Radio.defaultProps = {
   disabled: false,
   id: null,
   secondary: null,
-  value: null,
```
domhorvathtw: same comment as above

docs/utils/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as generateInput } from './generateInput';
```
domhorvathtw: Cool, like that we follow the same pattern and imports statements will be cleaner

docs/utils/generateInput/generatInput.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { Select, Checkbox } from '../../../src';
+
+const generateInput = (knobs, componentThis) => {
```
domhorvathtw: So when someone will create the docs for a component, he / she can create an array of knobs, and this function will generate them? 
Can we add an example in the description maybe? Would be nice to see it in action.

docs/utils/generateDocs/generateDocs.js
```diff
@@ -0,0 +1,28 @@
...
+  const docs = Object.values(knobs)
+    .map(knob =>
+      knob
+        .reduce((acc, curr) => {
+          const value =
+            componentThis.state[curr.state] && componentThis.state[curr.state].value
+              ? componentThis.state[curr.state].value
+              : componentThis.state[curr.state];
+          acc.push(`${curr.state} = {"${value || ''}"}`);
```
domhorvathtw: curr.state is the prop of the React component? If so, Can we stick to the naming convention and use `propName` or something?

does this work with all the different types for properties?
```
propertyOne="string"
propertyTwo={this.handleSomething}
propertyThree={false}
propertyThree={[1, 2, 3]}
```
etc.

src/common/RadioButton/RadioButton.js
andreapiras: no is not , it's just a quick fix for a console error value cannot be null. But I can remove it from this Pr

docs/utils/generateDocs/generateDocs.js
andreapiras: This are the single state properties of the Docs and they can also be the props of the component but not necessary that's why I rather keep them called state so we keep the context which is the docs

andreapiras:
```propertyOne="string"
propertyTwo={this.handleSomething}
propertyThree={false}
propertyThree={[1, 2, 3]}```
I added extraPropsDocs for the docs that we need but that are not Docs state

File: src/common/RadioButton/RadioButton.js (hunk 2 above)
Comment (domhorvathtw) [reply src/common/RadioButton/RadioButton.js]: not a big deal, it can stay :)

File: docs/utils/generateCodeBlock/generateCodeBlock.js [hunk 7]
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+const generateCodeBlock = (componentName, knobs, componentThis, extraPropsDocs = ['']) => {
```
dozortsev: the name is too generic, what exactly block of code it generates? block of knobs?

src/radioGroup/RadioGroup.docs.js
```diff
@@ -53,49 +55,17 @@ const KNOBS = {
...
-  };
-
   handleOnChange = selectedValue => {
     this.setState({ selectedValue });
     // eslint-disable-next-line
     console.log('Option selected', selectedValue);
   };
 
   render() {
+    const extraPropsDocs = ["onChange={'console.log'}"];
```
dozortsev: it seems useless to to mention about `console.log` for callback props, more useful I see showing functional literal and what values will be passed in it `(event | value) => { .. }`

andreapiras: You can put any string you prefer here :)

docs/utils/generateCodeBlock/generateCodeBlock.js
domhorvathtw: I suggested the name, I think it describes well that it's going to be a block of code. In the case of the Docs it's always `JSX`. Open for alternatives :)

PR #152, opened 2019-05-17
Np 61 radio option hot fix

andreapiras:


src/radioOption/RadioOption.js
```diff
@@ -43,13 +44,15 @@ RadioOption.propTypes = {
...
   disabled: Types.bool,
+  value: Types.string,
 };
 
 RadioOption.defaultProps = {
   description: null,
   checked: false,
   complex: false,
   disabled: false,
+  value: '',
```
dozortsev: `null`?

dozortsev: we have `null` as default value many other components

andreapiras: you can't use null as value for input :) Currently, if you create an input like <input value={null} onChange={this.handleChange} />, the null value is a flag for React to treat this as an uncontrolled input, and a console warning is generated.

PR #154, opened 2019-05-17
NP-4 Add Chevron component [WIP]

domhorvathtw:
## Simple Chevron component
**Jira ticket**
https://transferwise.atlassian.net/browse/NP-4

### Description
Small chevron component to indicate direction of things and animate directional changes. Probably not a standalone component and will be used inside other (more complex) components.

### Why?
Currently we use the `chevron` glyph in many components but it's coming from Bootstrap (or from TW icons in some cases). The main reason to have this  small molecule is to help us move away from Bootstrap and progressively create / recreate React components with this small utility component. With its current props it aims to cover most of the use cases that could be required by a host component (e.g. rotating, flipping, etc.)

### Props
- **Orientation** - top | right | bottom | left
- **Size** - small | medium | large
- **Flip** - boolean (flip chevron's direction)
- **Animate** - use CSS transitions as animations

### Animation
The component supports two different ways of animating direction changes. The reason it's implemented like this is to support both animation types: `rotation` and `flipping`.

**animation-duration:** '300ms'

`Rotation animation`
rotate

`Flipping animation`
flip

### Using it for other components
When using it in other components, we can implement the logic how the chevron should behave and render based on the container component's state. For changing the appearance of the chevron just use its `chevron` class while writing CSS for styling the host component.

### Sizing
Currently there are probably no use cases where we would use a chevron in different sizes (small, large), but I added it to the component. The default (medium) size is a `24 x 24px` container since our default paragraph's line height is `24px`. Based on a quick look around, maybe a `16 x 16px` container would be more appropriate.

### Todo
- [x] Adjust stroke width
- [x] Change default size to `16x16px`
- [x] Add tests
- [x] Move styles to Neptune CSS 

## Chevrons in the wild
image

image

image

image

image

image

image

image

image

image

image

image

image


gusso:
There's a bug in Safari where the element jumps at the end of the animation, probably worth taking a look.

Kapture 2019-05-28 at 8 40 03

domhorvathtw:
The safari bug was the lack of having both transform values (transformY and rotate) in the default state. 

**Before**
```less
  &.top { transform: rotate(-180deg) }
  &.top.flip { transform: rotate(-180deg) translateY(-30%) }
```

**After**
```less
  &.top { transform: rotate(-180deg) translateY(0%) }
  &.top.flip { transform: rotate(-180deg) translateY(-30%) }
```

src/common/Chevron/Chevron.less
```diff
@@ -0,0 +1,43 @@
+@import '~bootstrap/less/variables';
```
andreapiras: Add this to neptune-css <3

domhorvathtw: 👌

oliverviljamaa:
Love how detailed you are in your PR descriptions and comments 😍

src/common/Chevron/Chevron.js
```diff
@@ -0,0 +1,66 @@
+import React, { Fragment } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+import './Chevron.less';
+
+export const Orientation = {
```
oliverviljamaa: Do we normally export these in `CAPITAL` or in `PascalCase`?

src/common/Chevron/Chevron.js
```diff
@@ -0,0 +1,66 @@
...
+
+const Chevron = ({ orientation, flip, size, animate }) => {
+  const containerSize = CONTAINER_SIZE[size];
+  const classes = classNames(orientation, size, {
+    flip,
+    animate,
+  });
+
+  return (
+    <Fragment>
```
oliverviljamaa: Just curious, why do we need the `Fragment`?

src/common/Chevron/Chevron.js
oliverviljamaa: Why I'm asking is I'm used to seeing them in `UPPERCASE`:
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

oliverviljamaa: Also the first notion of exported enums seems to be here, also using `UPPERCASE`:
https://github.com/transferwise/components/blob/master/src/tooltip/Tooltip.js#L6

domhorvathtw: We had a similar discussion in the Button PR, and there we both liked the `"enum-like"` style for these names constants in `PascalCase`. I don't have a good reason the prefer one over the other. Maybe it's time to put these things ion a company wide styleguide?

src/common/Chevron/Chevron.js
domhorvathtw: my bad, fix it

src/common/Chevron/Chevron.js
oliverviljamaa: I would keep `Orientation` as `PascalCase` but the values in `UPPERCASE`, similarly to enums in Java. And you're correct, I approved the PR with `PascalCase` values as well 😅, totally my mistake!  What do you think is best?

domhorvathtw: I like the `UPPERCASE` style as well. If it was like that in other components, let's follow that. :D will update it now.

PR #155, opened 2019-05-20
Add popover bottom placements

phildobsontw:
Any popover that originates within 125 pixels of the left or right hand side of a 360px width viewport will overflow fairly significantly.

Old: (360px width viewport)
image

The fix: 
Adding two new bottom placements for the popover, BOTTOM_RIGHT and BOTTOM_LEFT.

These will be automatically selected if:
- Preferred position is left or right
- Placing the popover on both the left and the right would result in overflow
- Placing the popover below results in an overflow on one side (it will now therefore be shifted either left, or right).

Slight bit of refactoring to enable reuse of methods (and remove duplicate calculations).

New:
image

This isn't perfect - It's possible that with this logic up to 19px of the popover will overflow the left or right hand side. As the margin on the popover is 24px, this is a significant improvement as no text will be obscured.


src/popover/positioning.js
```diff
@@ -1,46 +1,60 @@
...
+  } else if (overflowsWithBottomPlacement.right && !overflowsWithBottomPlacement.left) {
+    return BOTTOM_LEFT;
+  }
+  return BOTTOM;
+}
+
+function calculateOverflows(popover, placement, documentBounds) {
+  const pos = getPopoverPosition(popover, placement);
+  return {
+    right: pos.right > documentBounds.right,
```
andreapiras: Maybe let's call it 'isOverflowRight' ?

src/popover/positioning.js
```diff
@@ -1,46 +1,60 @@
...
+  if (!rightPlacementOverflows.right && !leftPlacementOverflows.left) {
+    return preferredPlacement;
   }
-  if (overflowsRight && overflowsLeft) {
-    return BOTTOM;
+  if (rightPlacementOverflows.right && leftPlacementOverflows.left) {
+    return determineBestBottomPlacement(popover, documentBounds);
   }
-  if (overflowsRight && placement === RIGHT) {
+  if (rightPlacementOverflows.right && preferredPlacement === RIGHT) {
```
andreapiras: What if it does overflow left too?

src/popover/positioning.js
```diff
@@ -1,46 +1,60 @@
...
     return RIGHT_TOP;
   }
-  return placement;
+  return preferredPlacement;
+}
+
+function determineBestBottomPlacement(popover, documentBounds) {
+  const overflowsWithBottomPlacement = calculateOverflows(popover, BOTTOM, documentBounds);
+  if (overflowsWithBottomPlacement.left && !overflowsWithBottomPlacement.right) {
+    return BOTTOM_RIGHT;
```
andreapiras: Do we really need BOTTOM_LEFT AND BOTTOM_RIGHT?

src/popover/positioning.js
andreapiras: How do You know there s no overflow left

phildobsontw: I think we'd only be concerned with whether the popover if positioned on the left would overflow on the left if we were to actually intend to position it on the left. This should be covered on line 33 👍

src/popover/positioning.js
phildobsontw: `bottom-left` and `bottom-right` are classes both already available in our bootstrap CSS. Having both of these enables us to shift the popover either left or right based on which side it overflows when positioned at the bottom.

src/popover/positioning.js
phildobsontw: have gone with `overflowsRight` just to avoid it sounding like a method (might just be my sketchy java background 😉).

PR #156, opened 2019-05-21
Accordion

andreapiras:
**What?**
Adds accordion comoponent. You can play with it here: https://transferwise.github.io/components/branch/accordion/index.html

Animations were left out of this iteration intentionally to keep initial version as simple as possible. And I'm not fully sure if its worth adding, because it might cause performance issues on some cases. Might be the best to have prop that controls if animations are on or off. And IMO default should be off.
Not planning to continue with animations now.

**PS!** This PR description and code was done by @taaviaasver (Andrea created the PR).


src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,41 @@
...
+    return (
+      <Fragment>
+        {/* eslint-disable-next-line jsx-a11y/label-has-for */}
+        <label onClick={this.toggleIsOpen} className="accordion-item decision p-a-0">
+          <div className="media p-y-3">
+            <div className="media-body">
+              <h5>{title}</h5>
+            </div>
+            <div className="media-right">
+              <i className={`caret p-b-0 ${isOpen ? 'rotate180' : ''}`} />
```
andreapiras: Please replace with chevron

gusso: Resolved with 33da611

domhorvathtw: Cool!

src/accordion/AccordionItem/AccordionItem.less
```diff
@@ -0,0 +1,6 @@
+.accordion-item {
```
domhorvathtw: Do part of the styles come from Bootstrap? If not and it's all custom, do you think we should move it to Neptune CSS?

src/accordion/Accordion.less
```diff
@@ -3,4 +3,7 @@
   &:hover {
     background-color: #fff !important; // overrides bootstrap rule for .decision
   }
+  .chevron {
+    vertical-align: middle;
```
domhorvathtw: What do you think about using flexbox here? 
I'm trying to use it everywhere I can, but I might be biased :) 
https://philipwalton.github.io/solved-by-flexbox/demos/vertical-centering/

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,42 @@
...
+    return (
+      <Fragment>
+        {/* eslint-disable-next-line jsx-a11y/label-has-for */}
+        <label onClick={this.toggleIsOpen} className="accordion-item decision p-a-0">
+          <div className="media p-y-3">
+            <div className="media-body">
+              <h5>{title}</h5>
+            </div>
+            <div className="media-right media-middle">
+              <Chevron size="sm" orientation={isOpen ? 'top' : 'bottom'} animate />
```
domhorvathtw: Do we want to use the `flip` option here? It feels like when the change in the direction is 180 degrees, the flipping animation is more appropriate maybe?

src/accordion/AccordionItem/AccordionItem.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  it('displays the title', () => {
+    expect(component.find('h5').text()).toBe('Hello');
+  });
+
+  it('has content hidden by default', () => {
+    expect(component.instance().state.isOpen).toBe(false);
+    expect(getContent().text()).not.toContain('world!');
+  });
+
+  it('displays the content when clicked', () => {
```
StevePole: Given several of these 'click', I wonder if we could group the behaviour in another describe:

describe('when clicked', () => {
  beforeEach(() => {
    component.find('label').simulate('click');
  });
  it('should display the content (html element)', () => { ... }
  it('should show the chevron pointed up' () => { ... }
  ...
});

src/accordion/AccordionItem/AccordionItem.js
gusso: Oh yea sorry, that was me, we should definitely do that.

src/accordion/Accordion.spec.js
```diff
@@ -0,0 +1,22 @@
...
+        title: 'a',
+        content: 'b',
+      },
+      {
+        title: 'c',
+        content: <h1>I'm a h1 element</h1>,
+      },
+    ];
+    const component = shallow(<Accordion items={items} />);
+    expect(component.find(AccordionItem)).toHaveLength(2);
```
dozortsev: I think we also should have a test with empty array and when we don't specify the `items` prop at all

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,42 @@
...
+  static propTypes = {
+    title: Types.string.isRequired,
+    content: Types.node.isRequired,
+  };
+
+  state = {
+    isOpen: false,
+  };
+
+  toggleIsOpen = () => {
```
dozortsev: can be just `toggle`

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,42 @@
+import React, { PureComponent, Fragment } from 'react';
+import Types from 'prop-types';
+import Chevron from '../../common/Chevron';
+
+class AccordionItem extends PureComponent {
+  static propTypes = {
+    title: Types.string.isRequired,
+    content: Types.node.isRequired,
```
dozortsev: I would call it `children` because at some case we might start using it in way that it contains more than just test e.g test with some UI with elements like buttons, images, checkboxes etc. 

"content" is more about simple text.

src/accordion/AccordionItem/AccordionItem.js
dozortsev: here for `size="sm"` I would use [`Sizes.SMALL`](https://github.com/transferwise/components/blob/306040664430ad51a556b9ffbf46379d62274ee6/src/common/sizes.js#L3) and for `orientation` _Chevron_ has a [constant](https://github.com/transferwise/components/blob/abdce09dd035787946833bd84640cfa2544c0932/src/common/Chevron/Chevron.js#L6).

src/accordion/AccordionItem/AccordionItem.js
taaviaasver: Calling it `children` might be confusing - it implies that it should be used like that:
```
<AccordionItem>
  <h1>This is a child</h1>
</AccordionItem>
```

Or maybe its good?! I'm not sure..

src/accordion/Accordion.docs.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+];
+
+const AccordionDocs = () => (
+  <div className="container">
+    <section className="section">
+      <div className="row m-t-4">
+        <div className="col-md-6">
+          <h2>Accordion</h2>
+          <p>[:||||||||||||:]</p>
```
oliverviljamaa: :noice:

src/accordion/Accordion.spec.js
oliverviljamaa: I'd also test that each of the `AccordionItem`s gets the `title` as `key` and `title`, and `content` as `content`. WDYT?

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: I agree with @taaviaasver here, best to avoid clashing with the default `children` prop 😊

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,42 @@
...
+  toggleIsOpen = () => {
+    const { isOpen } = this.state;
+    this.setState({ isOpen: !isOpen });
+  };
+
+  render() {
+    const { title, content } = this.props;
+    const { isOpen } = this.state;
+    return (
+      <Fragment>
```
oliverviljamaa: Do we need the `Fragment` just for the ESLint disable comment?

taaviaasver: yes, any better alternatives?

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,42 @@
+import React, { PureComponent, Fragment } from 'react';
+import Types from 'prop-types';
+import Chevron from '../../common/Chevron';
+
+class AccordionItem extends PureComponent {
+  static propTypes = {
+    title: Types.string.isRequired,
```
oliverviljamaa: That probably wouldn't work with `<Message>key.here</Message>`, would it, so I believe we should have `node` here as well. Correct me if I'm wrong!

taaviaasver: My initial plan was to have it as node. Then I wanted to use `title` as list item identifier (`key` prop) and switched to `string`.

Any ideas on how to handle list item `key` if its a `node`?
Maybe using iteration `index` as key is fine for this case. I'm not sure. Or I could force consumer to specify `key` for every `item`.

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: Do we have a reasonable way to add the `htmlFor`? Semantically, it feels we may want a hidden checkbox?

src/accordion/AccordionItem/AccordionItem.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  beforeEach(() => {
+    component = shallow(<AccordionItem title="Hello" content="world!" />);
+  });
+
+  it('displays the title', () => {
+    expect(component.find('h5').text()).toBe('Hello');
+  });
+
+  it('has content hidden by default', () => {
+    expect(component.instance().state.isOpen).toBe(false);
```
oliverviljamaa: Do you think it's necessary to test the component internals here? I'd only keep the second assertion 🙃

src/accordion/AccordionItem/AccordionItem.spec.js
oliverviljamaa: I agree this approach is very BDD, but it has the downside of not grouping the checks for the same part of the component together, which can make it a bit tough to follow. 🤷‍♂

src/accordion/AccordionItem/AccordionItem.js
dozortsev: @taaviaasver  I actually would say that we should use like you mentioned https://github.com/transferwise/components/pull/156#discussion_r292405391

src/accordion/AccordionItem/AccordionItem.spec.js
oliverviljamaa: In the current case, I would even do a composite test (as the toggling test would fail anyway if the content was shown by default):
```js
it('has content hidden by default and togglable on label click', () => {
  expect(contentShown()).toBe(false);
  clickLabel();
  expect(contentShown()).toBe(true);
  clickLabel();
  expect(contentShown()).toBe(false);
});
```

src/accordion/AccordionItem/AccordionItem.js
dozortsev: check how it's implemented [here](https://polaris.shopify.com/components/behavior/collapsible) the "content" is in `children` prop

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: Honestly, I'd use `index` for now, and wouldn't burden the consumer. We could also do something like:
```js
... key={getKeyForItem(item)} ...

function getKeyForItem({ title }) {
  if (typeof title === 'string') {
    return title;
  } else if (title.props && typeof title.props.children === 'string') {
    return title.props.children;
  }
  throw new Error('Accordion prop title must be string or have a string child');
}
```
but that may be pushing it. Whatever you think is best 🙂

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: That's a good point, both of the approaches are fine. This one creates a discrepancy between the title and content, though, but it may be fine. What's our typical approach in this repo? `Modal`, `Popover`, `Tooltip`, etc.?

src/accordion/Accordion.spec.js
taaviaasver: Added test for empty array and for props (if these are passed forward to `AccordionItem`).

I did not add a test case for `items` being `undefined`, because `isRequired` got added to the prop-types. Are you ok with this @dozortsev?

oliverviljamaa: I'm ok with this, _but_ I have to admit – I'm not @dozortsev.

src/accordion/AccordionItem/AccordionItem.spec.js
taaviaasver: I like the last example by @oliverviljamaa.

Used that + did some other changes, I think it's easily readable now.

src/accordion/AccordionItem/AccordionItem.js
dozortsev: I would go with `node` for `title` prop and use `index#toString` in `key` attr, so that in _Accordion_ I would have something like following:

```html
<Fragment>{items.map((itemProps, index) => <AccordionItem key={index.toString()} {...itemProps} />)}</Fragment>
```

----

I'm not sure but afaik having `key={index.toString()}` is a bit sucks but I think it's good for now to at least to disable eslint, until Design System team will add some UUID lib for such things, make sense?)

src/accordion/AccordionItem/AccordionItem.js
dozortsev: > ..What's our typical approach in this repo?

Not sure but I think we don't have a convention for this, but yeah components like [_Alert_](https://github.com/transferwise/components/blob/82febb14b3dc86f4530f3a103e9b9b1d7b064ad9/src/alert/Alert.js#L30), [_Popover_](https://github.com/transferwise/components/blob/ecb89bcbc6b53c13498304598c7d4dbbde96bbe4/src/popover/Popover.js#L20),  [_Tooltip_](https://github.com/transferwise/components/blob/8910dcd82d9d1a136e68a023561ee2036a653984/src/tooltip/Tooltip.js#L69) (maybe few more) use `children` for "content".

src/common/Chevron/index.js
```diff
@@ -1 +1 @@
-export { default } from './Chevron';
+export { default, Orientation, Size } from './Chevron';
```
dozortsev: hmm, strange, how you get `Size` from `Chevron` folder? 🤔

src/accordion/AccordionItem/AccordionItem.js
taaviaasver: The snippet won't work if the element is more complex (multiple children, deeply nested).

I thought about doing smth like: 
```
import { renderToStaticMarkup } from 'react-dom/server';

...

function getKeyForItem({ title }) {
  return renderToStaticMarkup(title);
}
```

But I'm not so sure about performance.

**I think I'm going for the simplest solution, just using index for now**

src/accordion/AccordionItem/AccordionItem.js
dozortsev: I think it's not a big deal for now, I'd go with this approach, DS team will review and fix & improve all a11y things at some points, it's in our plans 😉

src/accordion/Accordion.spec.js
dozortsev: yes, looks good

src/common/Chevron/index.js
taaviaasver: `Chevron.js` exports it. `from './Chevron'` means `Chevron.js` 

```
export const Size = {
  SMALL: 'sm',
  MEDIUM: 'md',
  LARGE: 'lg',
};
```

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: Yup, I'd go with index as well.

src/accordion/AccordionItem/AccordionItem.js
taaviaasver: I think it doesn't make much sense for this case:

- we are talking about "internal" component `AccodionItem`. Consumers will use `Accordion`, which accepts array as prop - and IMO its clearer when array item prop is named `content`
- I don't like the discrepancy between the title and content that @oliverviljamaa pointed out
- for `Alert` and some others it totally makes sense

So I'd keep it as it is for now.

src/accordion/Accordion.js
```diff
@@ -0,0 +1,24 @@
+import React, { Fragment } from 'react';
+import Types from 'prop-types';
+
+import AccordionItem from './AccordionItem';
+
+import './Accordion.less';
+
+const Accordion = ({ items }) => (
```
oliverviljamaa: I don't like the idea of adding the `Fragment` solely for the comment, what do you think about this:
```js
/* eslint-disable react/no-array-index-key */
const Accordion = ({ items }) =>
  items.map((item, index) => <AccordionItem key={index} {...item} />);
/* eslint-enable react/no-array-index-key */
```

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: That's true, I didn't even think of the fact that it's not a public component.

src/accordion/AccordionItem/AccordionItem.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    clickLabel();
+    expect(getContent().contains(htmlElement)).toBe(true);
+  });
+
+  const getTitle = () => component.find('h5');
+  const getContent = () => component.find('label div').last();
+  const getChevron = () => component.find(Chevron);
+  const contentShown = () => getContent().contains(props.content);
+  const clickLabel = () => component.find('label').simulate('click');
+});
```
oliverviljamaa: I love how these tests turned out.

src/accordion/AccordionItem/AccordionItem.js
dozortsev: all right

src/accordion/Accordion.js
dozortsev: sorry am I missing something? why not just use `index.toString`? it should fix eslint

```
items.map((item, index) => <AccordionItem key={index.toString()} {...item} />);
```

oliverviljamaa: It's a workaround, not a solution, though, and I'd rather be explicit about breaking a rule. Even as a string, the index changes when an element is added somewhere in the array, creating the exact same inefficiency for which the rule was created.

oliverviljamaa: What do you think, @dozortsev? 🙃

dozortsev: I mean I know it's sucks, I think using UUID generator should solve the issue, right? we'll add it to this repo and fix as comments as such bad workarounds, deal? :)

dozortsev: this thing is not a big deal, and will be refactored at some point, @taaviaasver go with Oliver's approach if you prefer it more 🙂

taaviaasver: Went for your's, then switched back. I'd now keep it as it is 🙂

oliverviljamaa: Yup, @dozortsev, agreed!
It'll be even easier to add the UUIDs when we can find them by the ESLint disabled rule names :dab:

PR #158, opened 2019-05-22
Converted InputWithPattern to HOC

andreapiras:
**Jira ticket**
https://transferwise.atlassian.net/browse/NP-68

**Description**
Text format already implemented as regular component as been converted into HOC, as conceptually it  's more correct, and also it can be applied to input or textarea if necessary.


src/HOC/withTextFormat/withTextFormat.spec.js
```diff
@@ -129,18 +117,15 @@ describe('InputWithTextFormat', () => {
...
     it(`when entered a char before a symbol`, () => {
-      component.setState({
-        triggerEvent,
-      });
       component.setState({ selectionStart: 2, selectionEnd: 2 });
       component.simulate('change', { target: { value: '13-@' } });
       jest.runAllTimers();
       expect(component.state().selectionStart).toBe(4);
     });
-
```
StevePole: ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/components",
-  "version": "6.2.0",
+  "version": "7.0.0",
```
StevePole: Is this a breaking change?

andreapiras: Yes as I removed a component so if someone is using it it won't be there anymore

PR #165, opened 2019-06-05
Add strictMath to Less loader

gusso:
## Why?
Our version of the Less loader fails to process a CSS file that has custom properties inside `calc()` functions, because it only expects numerical values and Less variables. This fails the build process when importing some [files](https://github.com/transferwise/neptune-css/blob/master/src/less/flex.less) from Neptune CSS.

With [`strictMath`](http://lesscss.org/usage/#less-options-math) on, all math operations need to be wrapped in parenthesis to be processed by Less.

Bootstrap 3 [adheres to `strictMath`](https://github.com/twbs/bootstrap/pull/11807) by wrapping all maths in parenthesis, so this doesn't affect our current CSS base.

## Couldn't we separate the CSS and Less rules on webpack?
Probably? If we have separate rules for `css` and `less` files, the Less loader wouldn't complain about the imported files from Neptune CSS, because they all come compiled.

But `ExtractTextPlugin` is used to create a single CSS file in production, and I'm honestly unsure how to break this apart in two rules while keeping the result.

https://github.com/transferwise/components/blob/41144358aef9d8bd0267bdc245b38ef112c1f38e/config/webpack.config.prod.js#L138-L145

We'd also potentially run into the same problem when/if we create custom properties inside `calc()` functions in Less files. So as long as we're keeping our Bootstrap build process with this version of the Less loader, `strictMath` looks like the most reasonable solution to me.

## Changes
- Update the 3 webpack configs to use `strictMath`


codewhisperer:
LGTM, I don't think there is a good way to have separate rules but one CSS bundle

PR #170, opened 2019-06-20
Np 9 drawer

andreapiras:
# Drawer Component

## Jira 

https://transferwise.atlassian.net/jira/software/projects/DS/boards/277?selectedIssue=NP-9

## Demo

https://transferwise.github.io/components/branch/NP-9-drawer/index.html

### Description

Drawer component allows to slide content on faded background. Drawer component has been build combining two other components. 
1) Dimmer (faded background potentially to be shared with Modal)
2) Slider (simple slider)

A fourth component has been implemented: Sticky

Sticky is a wrapper for Slider and allows to have a sticky content at top/bottom/left/right page position.


src/dimmer/Dimmer.docs.js
```diff
@@ -0,0 +1,69 @@
...
+import '@transferwise/neptune-css/dist/css/flex.css';
+import classNames from 'classnames';
+
+import Dimmer from './';
+import { generateCodeBlock, generateInput, generateState } from '../../docs/utils';
+import { Button } from '../index';
+
+const KNOBS = {
+  knobs: [
+    { type: 'checkbox', label: 'Open', state: 'open', defaultState: false },
```
StevePole: How come we haven't changed this format yet, thought you were going to?  Job is just getting bigger.

src/dimmer/Dimmer.js
```diff
@@ -0,0 +1,80 @@
...
+import Types from 'prop-types';
+import classNames from 'classnames';
+import CSSTransition from 'react-transition-group/CSSTransition';
+import { createPortal } from 'react-dom';
+import '@transferwise/neptune-css/dist/css/dimmer.css';
+import { addModalOpenBodyClass, removeModalOpenBodyClass } from '../common';
+
+import KEY_CODES from '../common/keyCodes';
+
+export const EXIT_ANIMATION = 600;
```
StevePole: This is too slow.

src/dimmer/Dimmer.js
```diff
@@ -0,0 +1,80 @@
...
+
+const Dimmer = ({ open, children, onClose, fadeContentOnExit, fadeContentOnEnter }) => {
```
rs6g10: Maybe you can have a look at how Modal has done and maybe create something common so it's shared amongst components?

src/dimmer/Dimmer.js
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+    </CSSTransition>
+  );
+};
+
+Dimmer.propTypes = {
+  open: Types.bool,
+  children: Types.node,
+  onClose: Types.func,
+  fadeContentOnExit: Types.bool,
```
StevePole: Why should this be configurable?  Shouldn't we just make a decision about what we want?  Content should either always fade in or not.  (Same goes for out).

src/dimmer/Dimmer.js
```diff
@@ -0,0 +1,80 @@
...
+const Dimmer = ({ open, children, onClose, fadeContentOnExit, fadeContentOnEnter }) => {
+  const handleOnKeyDown = event => {
+    if (event && (event.keyCode === KEY_CODES.ESCAPE || event.key === 'Escape')) {
+      handleOnClose(event);
+    }
+  };
+
+  const handleOnEnter = () => {
+    addModalOpenBodyClass();
+    document.addEventListener('keydown', handleOnKeyDown);
```
rs6g10: Might wanna consider thinking about the mounting point wherein document could be `undefined`.

src/slider/Slider.docs.js
```diff
@@ -0,0 +1,109 @@
...
+                and height. Additional classes can be passed to show the slider full screen or with
+                a border. In order for those classes to be applied CSStransition needs to be
+                triggered
+              </span>
+              {generateCodeBlock('Slider', KNOBS, this)}
+            </div>
+            <div className="col-md-6">
+              <div
+                className="col-xs-12 m-b-5 p-x-0"
+                style={{ height: '300px', overflow: 'hidden', border: '1px solid #d3d5d8' }}
```
StevePole: ?

src/drawer/Drawer.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+import '@transferwise/neptune-css/dist/css/drawer.css';
```
StevePole: The animation is weird, and too slow.

src/drawer/Drawer.js
```diff
@@ -0,0 +1,104 @@
...
+  onClose: Types.func,
+  headerTitle: Types.string,
+  spacing: Types.string,
+  footerContent: Types.node,
+};
+
+Drawer.defaultProps = {
+  open: false,
+  children: null,
+  fadeContentOnExit: false,
```
StevePole: Again, not necessary to provide this option.

src/dimmer/Dimmer.js
rs6g10: You have, my bad.

src/slider/Slider.js
```diff
@@ -0,0 +1,44 @@
...
+    unmountOnExit
+  >
+    <div className="slider">{children}</div>
+  </CSSTransition>
+);
+
+Slider.propTypes = {
+  open: Types.bool,
+  sliderPositionFixed: Types.bool,
+  showSliderBorder: Types.bool,
```
StevePole: Should this be optional?  Why don't they always look the same?

src/sticky/Sticky.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import Types from 'prop-types';
+
+import { Slider } from '..';
+
+export const EXIT_ANIMATION = 600;
+
+const Sticky = ({ open, position, children }) => (
```
StevePole: A lot of overlap between this and the slider.  How is a sticky right different to a slider right?  Why do we need both?

mikknarusk:
@andreapiras what's the rationale for using a different close icon for the Drawer component? Looks out of place in TW context imo

src/dimmer/Dimmer.js
domhorvathtw: 300 ms is usually the "rule of thumb" in web UIs, or considered as "normal"

https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9

src/dimmer/Dimmer.docs.js
andreapiras: Hey Steve I started this ticket before we merged the other , next one I'm going to change the format promise

src/dimmer/Dimmer.spec.js
```diff
@@ -0,0 +1,152 @@
...
+    component.setProps({ fadeContentOnEnter: true });
+    const cssTransition = component.find('CSSTransition');
+    expect(cssTransition.prop('classNames')).toEqual({
+      enter: 'dimmer--enter-fade',
+      enterDone: 'dimmer--enter-done dimmer--enter-fade',
+      exit: 'dimmer--exit',
+    });
+  });
+
+  it.only('fade content on exit if fadeContentOnExit is true', () => {
```
mikknarusk: only

andreapiras:
> Do all of these need to be documented? Should you ever use a slider without a dimmer? On the other hand I can see that dimmer could be shared by modals...

I think the usage of these component should live in the [neptune-docs ](https://transferwise.github.io/neptune-docs/). I think designers should decide whether or not a slider is usable on its own. At the moment from a developer point of view is reused to create the Sticky component. Regarding the Dimmer ther reason why I implemented as a separate component is because is going to be shared with Modal in a separate ticket.

src/dimmer/Dimmer.js
andreapiras: You mean for SSR issues?

src/dimmer/Dimmer.js
andreapiras: The reason why is there it is because I can see the modal content is fading while we don't want for example to fadeOut/In the drawer. If the animation of the content is for some reason slower than the Dimmer or the content is going to disappear with a "jump" because the component gets unmounted on exit

src/dimmer/Dimmer.js
andreapiras: > Maybe you can have a look at how Modal has done and maybe create something common so it's shared amongst components?

Hey Rahul, yes that's the plan. I extracted into its own component as it's going to be shared with Modal :)

src/slider/Slider.docs.js
andreapiras: This is just for docs, is basically creating a box where you can see the slider sliding.

src/dimmer/Dimmer.js
gusso: Agreed. I just suggested a new curve and 350ms for the time. I agree with the rule of thumb, but the mass on the drawer is very heavy, so it can't be too fast.

We did some experimentation with changing the opacity of the Dimmer that made the Drawer way less daunting, but let's tackle it after we evaluate if the popups would handle a more subtle Dimmer.

src/drawer/Drawer.js
gusso: There's a new one coming.

src/sticky/Sticky.js
gusso: This came out of [this conversation](https://transferwise.slack.com/archives/C2XG7R3EZ/p1560856627000400).

It's mostly about naming, Sticky would be a named instance of a Slide with certain props set. It felt uncomfortable to refer to a Slide every time a Sticky is used, as in it could come with the expectation it would slide eventually.

src/drawer/Drawer.js
andreapiras: As per commented on the ticket this will be updated to ```cubic-bezier(0, 0.94, 0.62, 1) 350ms```

src/dimmer/Dimmer.js
andreapiras: This comment was relative to 
```export default props =>
  typeof document !== 'undefined' ? createPortal(<Dimmer {...props} />, document.body) : '';```

which is at the end of the file

File: src/dimmer/Dimmer.js (hunk 2 above)
Comment (andreapiras) [reply src/dimmer/Dimmer.js]: As per commented on the ticket this will be updated to cubic-bezier(0, 0.94, 0.62, 1) 350ms

File: src/slider/Slider.js (hunk 9 above)
Comment (andreapiras) [reply src/slider/Slider.js]: The drawer doesn't have a border because of the dimmer while the sticky has

**Comment by andreapiras** [comment]
> @andreapiras what's the rationale for using a different close icon for the Drawer component? Looks out of place in TW context imo

@mikknarusk I've checked the home page and the cross there is a styled button. I'm currently using a tw-icon. To be honest it feels that using icons would bring much more consistency than using an ad hoc styled button. As far as I'm aware an external is looking to redesign our icons so probably would be better to stick to them as much as we can now.

File: src/dimmer/Dimmer.js (hunk 4 above)
Comment (StevePole) [reply src/dimmer/Dimmer.js]: Shouldn't modals and drawers feel like part of the same _system_, and behave in largely the same way?

File: src/sticky/Sticky.js (hunk 10 above)
Comment (StevePole) [reply src/sticky/Sticky.js]: From my reading it seems like the intended use case for a sticky is either top or bottom, while for a slide it's left or right.  Why not limit them to how they're supposed to be used?  There would then be a clear difference leading to people choosing the right one.  Then, if we decide to iterate on them later, it will likely be much easier as they have a more constrained use case.

**Comment by StevePole** [comment]
> @andreapiras what's the rationale for using a different close icon for the Drawer component? Looks out of place in TW context imo

Do we have a 'close' component yet?  We probably should.

**Comment by StevePole** [comment]
> > @andreapiras what's the rationale for using a different close icon for the Drawer component? Looks out of place in TW context imo
> 
> @mikknarusk I've checked the home page and the cross there is a styled button. I'm currently using a tw-icon. To be honest it feels that using icons would bring much more consistency than using an ad hoc styled button. As far as I'm aware an external is looking to redesign our icons so probably would be better to stick to them as much as we can now.

We shouldn't need to import our whole icon set to power basic UI.  Fine with using the same glyph of course.

File: src/sticky/Sticky.js (hunk 10 above)
Comment (StevePole) [reply src/sticky/Sticky.js]: Think this would also resolve stuff like this:
> The drawer doesn't have a border because of the dimmer while the sticky has

**Comment by andreapiras** [comment]
> > @andreapiras what's the rationale for using a different close icon for the Drawer component? Looks out of place in TW context imo
> 
> Do we have a 'close' component yet? We probably should.

@StevePole No we don't yet. Yes I think we should have one as we have the Chevron too

File: src/sticky/Sticky.js (hunk 10 above)
Comment (andreapiras) [reply src/sticky/Sticky.js]: @StevePole Slider is used to build both Drawer which now is limited to 
```position: Types.oneOf(['left', 'right']),```
and Sticky which now is 
```position: Types.oneOf(['top', 'bottom'])```

I think Slider should still have the optional border to cover both cases for Drawer (without border) and Sticky (with border)

File: src/dimmer/Dimmer.js (hunk 4 above)
Comment (gusso) [reply src/dimmer/Dimmer.js]: Right, but we went this route because the drawer has the requirement of not fading in and out as it slides. It should act like a piece of paper peaking into the page, and the fading reduces the effect.

I agree this diverges from the modal, but it didn't seem like a problem to me because the modal doesn't touch the edges and has the rounded corners, so it's a different kind of plane that flies in and out rather than slide, the fading extends the sense of motion.

We want to look into reducing the strength of the dimmer and for that we'll need to look into how this affects modals as well, so I could put these into a ticket to be looked separately as adjustments to Modal.

**Comment by anttio** [comment]
On mobile browser the drawer component top and bottom positions doesn't work properly. The drawer is too tall and gets cut and hides the CTA or the close button.

**Comment by andreapiras** [comment]
> On mobile browser the drawer component top and bottom positions doesn't work properly. The drawer is too tall and gets cut and hides the CTA or the close button.

Hey, sorry I forgot to remove those options from the docs. The drawer will only be open on Left or Right position. The Sticky component will always be opened from top or bottom

**Comment by gusso** [comment]
> We shouldn't need to import our whole icon set to power basic UI. Fine with using the same glyph of course.

@StevePole What do you mean? If the Icons repo is already a dependency, and the output is a single inline SVG, why shouldn't we do this?

**Comment by StevePole** [comment]
>@StevePole What do you mean? If the Icons repo is already a dependency, and the output is a single inline SVG, why shouldn't we do this?

Is icons already a dependency?  I don't know of any reason for it to be, other than the chevron and the close.  Don't think we should import an entire icon set for those two glyphs.

**Comment by andreapiras** [comment]
> > @StevePole What do you mean? If the Icons repo is already a dependency, and the output is a single inline SVG, why shouldn't we do this?
> 
> Is icons already a dependency? I don't know of any reason for it to be, other than the chevron and the close. Don't think we should import an entire icon set for those two glyphs.

Hey @Steve yeah the icons is a dependency and it is used in multiple places across the library. To be honest it seems legit to me that we use the icons as dependency but if we intend to change this maybe we can address the change in a separate ticket?

**Comment by StevePole** [comment]
What do you think about removing the docs for the 'slider'?  I find it a bit confusing.  Also not entirely sure about the name.  There's a pretty common pattern of this sort of thing being called a 'slider':

https://jqueryui.com/slider/

We don't currently use anything like that, but would try and avoid using the same name as a common component.

**Comment by andreapiras** [comment]
> What do you think about removing the docs for the 'slider'? I find it a bit confusing. Also not entirely sure about the name. There's a pretty common pattern of this sort of thing being called a 'slider':
> 
> https://jqueryui.com/slider/
> 
> We don't currently use anything like that, but would try and avoid using the same name as a common component.

Yeah sure we can remove it. I put it there so people know that it exists but is not crucial to have it there since it probably won't be used on its own

File: src/drawer/Drawer.js [hunk 12]
```diff
@@ -0,0 +1,93 @@
...
+  );
+};
+
+Drawer.propTypes = {
+  open: Types.bool,
+  position: Types.oneOf(['left', 'right']),
+  children: Types.node,
+  onClose: Types.func,
+  headerTitle: Types.string,
+  spacing: Types.string,
```
StevePole: This is undocumented.  I'd also say it's unnecessary.  We should decide what padding drawers should have and stick to it.  It would also be great if that was responsive so we don't have massive padding on portrait phones.

src/drawer/Drawer.js
```diff
@@ -0,0 +1,93 @@
...
+            >
+              <Close />
+            </div>
+            {headerTitle && <h3 className="drawer-header--title m-t-1 m-l-2">{headerTitle}</h3>}
+          </div>
+          {children && (
+            <div
+              className={classNames('drawer-content ', {
+                'p-y-5': headerTitle,
+                'p-x-5': spacing === 'lg',
```
StevePole: Feels a bit weird using utility classes, drawers should just have their own styles, you shouldn't need to apply an additional class.

src/drawer/Drawer.js
andreapiras: Hey @StevePole fixed that :)

gusso:
Screenshot 2019-07-05 at 1 34 58 pm
On mobile the title is not vertically centered within the header, and the top and bottom margins of the footer button area are sometimes too small (should be 12px).

andreapiras:
> Screenshot 2019-07-05 at 1 34 58 pm
> 
> On mobile the title is not vertically centered within the header, and the top and bottom margins of the footer button area are sometimes too small (should be 12px).
@gusso fixed bug on safari landscape

andreapiras:
> > Screenshot 2019-07-05 at 1 34 58 pm
> > On mobile the title is not vertically centered within the header, and the top and bottom margins of the footer button area are sometimes too small (should be 12px).
> > @gusso fixed bug on safari landscape

PR #174, opened 2019-07-03
Hide Modal divider if title is not provided

gusso:
## Description
Hiding all dividers was already possible by passing `.no-divider` as a className to the component, but this removes that necessity while maintaining the divider on the footer when there is one. It also replicates the prop behaviour of [Drawer](https://github.com/transferwise/components/pull/170).

## Context
https://transferwise.slack.com/archives/C2XG7R3EZ/p1562083990071700

## Demo
https://transferwise.github.io/components/branch/remove-modal-header-line/index.html

## Example
Approve component


src/modal/ModalDialog/ModalDialog.js
```diff
@@ -50,7 +50,7 @@ export default class ModalDialog extends Component {
       >
         <div className={classes}>
           <div className="modal-content">
-            <div className="modal-header">
+            <div className={classNames('modal-header', { 'no-divider': !title })}>
```
oliverviljamaa: Can we have a test for this, please?

rs6g10:
What Oliver said

src/modal/ModalDialog/ModalDialog.spec.js
```diff
@@ -73,6 +73,20 @@ describe('ModalDialog', () => {
 
       expect(component).toMatchSnapshot();
     });
+
+    it('with header divider if title is provided', () => {
+      component.setProps({ title: 'Some title' });
+      const header = component.find('.modal-header');
+
+      expect(header.hasClass('no-divider')).toBe(false);
```
oliverviljamaa: It seems we're using snapshot tests in this file for other similar tests – as that's the case, I'd follow suite, and just do:
```js
expect(component).toMatchSnapshot();
```
here as well!

PR #177, opened 2019-07-05
Phone number input width flexibility

twrdjpalmer:
Currently the phone number input's layout makes it very difficult to use inside containers of varying size. The country code select can become unnecessarily large, or the number input impossibly small.

The change here is to remove the reliance on the bootstrap grid, and thus have greater control and flexibility over the component's layout

**Before**

Screenshot 2019-07-08 at 10 45 28

**After**

phone-number-after


src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,12 @@
...
+  display: flex;
+}
+
+.tw-telephone__country-select .btn-group.btn-block .btn {
+  width: 111px;
+}
+
+.tw-telephone__number-input {
+  width: 100%;
+  padding-left: 16px;
```
gusso: I worry that the difference between 16px and 24px (the regular space between fields) is not large enough to seem intentful, it could look weird in the middle of a form with a lot of 24px spaces. I understand the intent is bringing the fields closer to associate them though, but then 8px looks too tight.

The options I can think of are leaving the margins at 24px, or considering coding the adapted pattern that @miketwigg used on his design.

What do you think?

miketwigg: Hey, if we are talking about the space between fields. I think 24px could work.
I've mocked it up and looks good. I'd need Richard to check that this size fits in the borderless section on HP.

Screenshot 2019-07-05 at 13 47 30

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,12 @@
+.tw-telephone {
+  display: flex;
+}
+
+.tw-telephone__country-select .btn-group.btn-block .btn {
+  width: 111px;
```
domhorvathtw: Is there a specific reason for `111px` in width?
If we want to use our base unit (8px) we can go for 14x8px = 112px.

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,12 @@
+.tw-telephone {
```
domhorvathtw: Since you are using flex layout, you could / should use `flex-basis` and `flex-grow` for the flex items in the container, or just use the flex shorthand:
**Small example:**
[Codepen example](https://codepen.io/domhorvath/pen/WqgLqy)
```css
.flex1 { flex: 0 0 112px; }
.flex2 { flex: 1 1 } 
```

[More about the flex shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/flex)

src/phoneNumberInput/PhoneNumberInput.less
twrdjpalmer: 111px was the minimum width required to show the four digit country code. 112px sounds better 👍

src/phoneNumberInput/PhoneNumberInput.less
twrdjpalmer: Setting `flex-basis` didn't seem to work when I tried originally. Perhaps due to missing the accompanying `flex-grow`—I'll give it a whirl

twrdjpalmer: Turns out I slightly misunderstood how `flex-shrink` works. Thanks for getting me to look at it again 🙌

src/phoneNumberInput/PhoneNumberInput.less
twrdjpalmer: Gone for 15x instead of 14x to make sure it works in unforeseen edge cases

domhorvathtw: Cool, sounds good!

domhorvathtw:
LGTM, don't forget to update package.json version number and the changelog.md with a short description!

mikknarusk:
The width seems to be fixed, so can we check how it looks like in other languages with a longer placeholder text. Maybe the width should be dynamic
Screenshot 2019-07-08 at 11 04 01

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,19 @@
+@import '~bootstrap/less/variables';
+
+@four-digit-country-code-width: 120px;
+
+.tw-telephone {
+  display: flex;
```
mikknarusk: does this go through autoprefixer or should we provide predfixed css?

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.tw-telephone__country-select {
+  flex-basis: @four-digit-country-code-width;
+  flex-shrink: 0;
+}
+
+.tw-telephone__number-input {
+  flex-basis: auto;
```
mikknarusk: should be ablet to use `flex: auto 1 1;` here AFAIK

src/phoneNumberInput/PhoneNumberInput.docs.js
```diff
@@ -44,7 +44,7 @@ const KNOBS = {
       type: 'text',
       label: 'Placeholder',
       state: 'placeholder',
-      defaultState: 'placeholder',
+      defaultState: '12345678901234',
```
andreapiras: This will fill the input with +44 because the value provided is not a valid phone number. If you want the field to be populated correctly please add + .  +1234567890

src/phoneNumberInput/PhoneNumberInput.less
```diff
@@ -0,0 +1,19 @@
...
+  flex-grow: 1;
+  flex-shrink: 1;
+  margin-left: (@padding-base-vertical * 3);
```
andreapiras: uhmm this looks weird :) can we not use an utility class like m-l-3 ?

src/phoneNumberInput/PhoneNumberInput.less
twrdjpalmer: Habit from using css-in-js and libs which prefer zero short hand to avoid deterministic issues 😇

src/phoneNumberInput/PhoneNumberInput.less
twrdjpalmer: @mikknarusk quick test shows autoprefixer is there and working 👍

src/phoneNumberInput/PhoneNumberInput.docs.js
twrdjpalmer: @andreapiras further investigation: this is for the placeholder of the input only, changing it to +44... isn't quite right:

Screenshot 2019-07-08 at 15 36 31

twrdjpalmer:
> The width seems to be fixed, so can we check how it looks like in other languages with a longer placeholder text. Maybe the width should be dynamic

Got any ideas on the dynamic width? Making the input itself have dynamic width ends up in a really strange experience with the number input moving about as you choose different width country codes. 

It's worth noting that few, if any inputs I'm aware of size themselves to their placeholder text. Would be great to hear your thoughts on it @mikknarusk

mikknarusk:
> > The width seems to be fixed, so can we check how it looks like in other languages with a longer placeholder text. Maybe the width should be dynamic
> 
> Got any ideas on the dynamic width? Making the input itself have dynamic width ends up in a really strange experience with the number input moving about as you choose different width country codes.
> 
> It's worth noting that few, if any inputs I'm aware of size themselves to their placeholder text. Would be great to hear your thoughts on it @mikknarusk

Valid points.
Maybe just look through a couple of languages (German and Russian are usual suspects) and see if we have a problem at all or not

twrdjpalmer:
@mikknarusk luckily German is the same length, and Russian is `префикс` which fits just fine  🙌

PR #178, opened 2019-07-05
Typeahead bootstrap

gittymcgit:
Added typeahead component

### Main use-cases:
image
_business lookup_

image
_chips_

image
maxHeight on chips mode

Also can be used as multiselect, however by default it will allow user to select duplicate items in multiselect mode, and it is up to parent component to filter options accordingly to selected items.

Used debounce from lodash to debounce an event for requesting new items. Delay is configurable. min length for search query is also configurable 

*PS: tests are failing now, because it needs an updated version of neptune, for which I have to merge chips*


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v10.1.0
+## Add Typeahead componend
```
andreapiras: typo :)

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+    footer: null,
+    onSearch: () => {},
+    onChange: undefined,
+    validateChip: () => true,
+    filterBy: undefined,
+    size: 'md',
+    chipSeparators: [],
+  };
+
+  constructor() {
```
andreapiras: constructor is redundant

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
+import React, { Component } from 'react';
+import { AsyncTypeahead, Menu, MenuItem, Highlighter } from 'react-bootstrap-typeahead';
+import Types from 'prop-types';
+
+import 'react-bootstrap-typeahead/css/Typeahead.css';
+
+import Chip from './chip/Chip';
+import './Typeahead.less';
+
+const SIZE_MAPPING = {
```
andreapiras: can we reuse Sizes from common folder here ? :)

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+      const value = event.clipboardData.getData('text');
+      const regex = new RegExp(chipSeparators.join('|'));
+      if (value) {
+        const pastedChips = value.split(regex).map(label => ({ label: label.trim() }));
+        const selected = []
+          .concat(this.state.selected, pastedChips)
+          .filter(chip => chip && chip.label);
+
+        this.updateSelectedValue(selected);
+        event.preventDefault();
```
andreapiras: should this be the first line after function declaration?

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+  onPaste = event => {
+    const { allowNew, multiple, chipSeparators } = this.props;
+
+    if (allowNew && multiple && chipSeparators.length > 0) {
+      const value = event.clipboardData.getData('text');
+      const regex = new RegExp(chipSeparators.join('|'));
+      if (value) {
+        const pastedChips = value.split(regex).map(label => ({ label: label.trim() }));
+        const selected = []
+          .concat(this.state.selected, pastedChips)
```
andreapiras: Optional: this can be replaced with ```const selected = [...this.state.selected,...pastedChips]
          .filter(chip => chip && chip.label);```

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+          selected = [...this.state.selected];
+        }
+        const { value } = this.typeahead.getInstance().getInput();
+        if (value) {
+          selected.push({ label: this.typeahead.getInstance().getInput().value });
+          this.typeahead.getInstance().clear();
+          this.updateSelectedValue(selected);
+        }
+
+        event.preventDefault();
```
andreapiras: Again I will push this as high as possible.

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+    }
+  };
+
+  onKeyDown = event => {
+    const { showSuggestions, allowNew, multiple, chipSeparators } = this.props;
+
+    if (!showSuggestions && allowNew && multiple) {
+      if (['Enter', ...chipSeparators].includes(event.key)) {
+        let selected = [];
+        if (this.state.selected) {
```
andreapiras: This can be replaced with 
```
const selected = this.state.selected ? [...this.state.selected] : [];```

File: src/typeahead/Typeahead.js [hunk 8]
```diff
@@ -0,0 +1,272 @@
...
+          selected.push({ label: this.typeahead.getInstance().getInput().value });
+          this.typeahead.getInstance().clear();
+          this.updateSelectedValue(selected);
+        }
+
+        event.preventDefault();
+      }
+    }
+
+    if (this.props.multiple && event.key === 'Backspace') {
```
andreapiras: please use multiple here

andreapiras:
good job! MAybe we can also add chips in the docs since they can probably be reused and so other people knows they exists

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+
+    if (!showSuggestions && allowNew && multiple) {
+      if (['Enter', ...chipSeparators].includes(event.key)) {
+        let selected = [];
+        if (this.state.selected) {
+          selected = [...this.state.selected];
+        }
+        const { value } = this.typeahead.getInstance().getInput();
+        if (value) {
+          selected.push({ label: this.typeahead.getInstance().getInput().value });
```
andreapiras: why not label:value ?

src/typeahead/Typeahead.less
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  .form-control.rbt-input-multi {
+    float: none;
+    display: block;
+  }
+
+  .open > .dropdown-menu {
+    animation: none;
+    transition: 0.2s cubic-bezier(0.6, 0.2, 0.1, 1);
```
andreapiras: We are trying to keep consistencies in our animations. Can you please use 

```
@import "./variables/_transitions.less";
  transition: var(--transition-transform);
```

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,226 @@
...
+        {
+          label: 'Test',
+        },
+      ],
+    };
+    component = shallow(<Typeahead {...props} />);
+  });
+
+  it('renders icon', () => {
+    component.setProps({ icon: 'test' });
```
andreapiras: I'd also check before setProps
expect(component.find('.icon-test')).toHaveLength(0);

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
andreapiras: I would add this in the docs

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+import Types from 'prop-types';
+import { tokenContainer } from 'react-bootstrap-typeahead';
+import Close from '@transferwise/icons/react/close';
+
+import './Chip.less';
```
andreapiras: This should go in neptune css

src/typeahead/chip/Chip.js
andreapiras: I add this component in the root as it could be reused

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+
+const Chip = props => {
+  const { className, valid, size, option, onRemove } = props;
+  return (
+    <div
```
andreapiras: This can be re-written as ``` className={classNames(`chip rbt-token rbt-token-removeable `, className, {
        'has-error': !valid,
        [`chip-${size}`]: size,
      })}
    >```

src/typeahead/chip/Chip.less
```diff
@@ -0,0 +1,48 @@
+@import '~bootstrap/less/variables';
```
andreapiras: This should go in neptune-css

src/typeahead/chip/Chip.less
```diff
@@ -0,0 +1,48 @@
...
+
+.chip {
+  color: @brand-navy;
+  padding: @padding-base-vertical 42px @padding-base-vertical @padding-base-horizontal;
```
andreapiras: I'd use flex

src/typeahead/chip/Chip.less
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  &.has-error {
+    background: @brand-red;
+    color: @component-active-color;
+  }
+
+  &-sm {
+    padding: 5px 26px 3px 8px;
```
andreapiras: Those numbers looks a bit random. Normally we work with multiple of 8px

src/typeahead/chip/Chip.less
```diff
@@ -0,0 +1,48 @@
...
+    margin-bottom: 5px;
+    .remove-chip {
+      padding: 0 5px;
+    }
+  }
+}
+
+// .form-control-sm .chip {
```
andreapiras: Remove if not needed

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+      className={`chip rbt-token rbt-token-removeable chip-${size} ${
+        valid ? '' : 'has-error'
+      } ${className}`}
+    >
+      {option.label}
```
andreapiras: Pass directly the value to display

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+      <button className="btn-unstyled remove-chip" onClick={onRemove}>
+        <Close size="sm" />
+      </button>
+    </div>
+  );
+};
+
+Chip.propTypes = {
+  option: Types.objectOf(Types.string).isRequired,
```
andreapiras: This should be just a plain string

src/typeahead/chip/Chip.js
andreapiras: I would return null if option is null

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
+import React, { Component } from 'react';
+import { AsyncTypeahead, Menu, MenuItem, Highlighter } from 'react-bootstrap-typeahead';
```
andreapiras: This package introduce approximately 40kb 74.49kb -> 113.81kb. Why are we using vs custom solution

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+  size: Types.string,
+  valid: Types.bool,
+};
+
+Chip.defaultProps = {
+  className: '',
+  size: 'md',
+  valid: true,
+};
+export default tokenContainer(Chip);
```
andreapiras: Probably better to not add dependency from boostrap here

anttio:
typeahead-visual

Giving some general feedback of the UI visuals without seeing the original designs:
- Text in chips should be vertically aligned just like the close-icon is.
- Letter-spacing is different on chips vs. when writing on typeahead.
- Footer styles missing?
- On mobile there is this blurry subpixel border around the chips.
- Why are they red?

- Should the `icon` and `iconDivider` props be renamed to `addon` to be more aligned with what we use in BS?
- Do we really need the `iconDivider`?

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+const SIZE_MAPPING = {
+  sm: 'small',
+  md: undefined,
+  lg: 'large',
+};
+
+export default class Typeahead extends Component {
+  static propTypes = {
+    id: Types.string.isRequired,
+    options: Types.arrayOf(Types.object).isRequired,
```
dozortsev: I would use `Types.shape` here and describe it's properties:

```javascript
options: Types.arrayOf(Types.shape({
   value: Types.any,
   ...
})).isRequired,
```

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+    allowNew: Types.bool,
+    autoFocus: Types.bool,
+    clearable: Types.bool,
+    iconDivider: Types.bool,
+    isLoading: Types.bool,
+    multiple: Types.bool,
+    showSuggestions: Types.bool,
+    icon: Types.string,
+    maxHeight: Types.string,
```
dozortsev: I think here you should use predefined constants: 

```javascript
maxHeight: Types.oneOf([HeightType.NONE, HeightType.AUTO..])
```

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+    placeholder: Types.string,
+    footer: Types.node,
+    filterBy: Types.func,
+    validateChip: Types.func,
+    onSearch: Types.func,
+    onChange: Types.func,
+    chipSeparators: Types.arrayOf(Types.string),
+    size: Types.oneOf(['sm', 'md', 'lg']),
```
dozortsev: please use [`Sizes`](https://github.com/transferwise/components/blob/master/src/common/sizes.js) here

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+        event.preventDefault();
+      }
+    }
+  };
+
+  onKeyDown = event => {
+    const { showSuggestions, allowNew, multiple, chipSeparators } = this.props;
+
+    if (!showSuggestions && allowNew && multiple) {
+      if (['Enter', ...chipSeparators].includes(event.key)) {
```
dozortsev: I'm not sure we can use `.includes` function [it's not supported by IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility).

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+
+    return (
+      <Chip
+        disabled={disabled}
+        tabIndex={tabIndex}
+        key={idx}
+        valid={valid}
+        size={size}
+        option={option}
+        onRemove={onRemove}
```
dozortsev:
```html
<Chip
 disabled
 tabIndex
 key={idx}
 ...
/>
```

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+      />
+    );
+  };
+
+  renderItem = (item, menuProps, idx) => {
+    const { text } = menuProps;
+    const { label, note, secondary } = item;
+
+    return (
+      <MenuItem position={idx} option={item} key={idx} id={idx}>
```
dozortsev: why in `position` you pass `idx`? a bit confusing

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,272 @@
...
+        <Highlighter search={text.trim()}>{label}</Highlighter>
+        {note ? <span className="small m-l-1">{note}</span> : ''}
+        {secondary ? <span className="small text-ellipsis">{secondary}</span> : ''}
```
dozortsev: two above lines could slightly simplified with `&&`, e.g:

```javascript
{note && <span className="small m-l-1">{note}</span>}
```

src/typeahead/Typeahead.less
```diff
@@ -0,0 +1,142 @@
+@import '~bootstrap/less/variables';
+
+.typeahead {
```
dozortsev: @andreapiras @nabaroa shouldn't be this in neptune-css?

src/typeahead/chip/Chip.js
```diff
@@ -0,0 +1,39 @@
...
+import { tokenContainer } from 'react-bootstrap-typeahead';
+import Close from '@transferwise/icons/react/close';
+
+import './Chip.less';
+
+const Chip = props => {
+  const { className, valid, size, option, onRemove } = props;
+  return (
+    <div
+      className={`chip rbt-token rbt-token-removeable chip-${size} ${
```
dozortsev: use _classnames_ here

gittymcgit:
1) In that mode I think you can't select anything. You need either showSuggestions or allowNew. It's done this way so we could addNew options without hiding suggestions
2) Consumers of the component should style their footer the way they want it. It can either be a footer, or a pagination or whatever they want it to be. That's why we don't apply any style.

src/typeahead/input/Input.js
```diff
@@ -0,0 +1,166 @@
...
+    const { inputWidth } = this.state;
+    if (multiple) {
+      const hasValue = selected && selected.length > 0;
+      return (
+        <div
+          className="form-control typeahead__input-container"
+          onClick={() => {
+            this.inputRef.focus();
+          }}
+          style={maxHeight && { maxHeight }}
```
dozortsev: don't understand how this `maxHeight` change the input, could you please explain (and/or probably show a screenshot)?

and why do we need expose this as prop?

gittymcgit: Sure thing! I added a screenshot in the description 👆!

gittymcgit:
RE: adding items. It's how it works, except for TAB. I think we should not override tab there by default. But if you set allowNew to true and showSuggestions to false, it will add new chips as you press Enter or other buttons that you set in props as chipsSeparators. In the example in docs it is configured to create a new chip on Enter, space or comma

gittymcgit:
Oh I see what you mean now. I guess we could use enter here, when you don't have an option highlighted!

andreapiras:
Found issue on IE11 
image

anttio:
> I'm not sure how to select something if suggestions are disabled

I don't really understand why is there an option to disable suggestions in a Typeahead component? If they are disabled it's just a regular text input field.

gusso:
When `allowNew` and `multiple` are true and you press enter, it creates an empty chip, we probably don't want that.

gittymcgit:
fixed all of those above. Added a light animation to a chip

src/typeahead/Typeahead.less
andreapiras: I think it makes sense for it to live in neptune css

gusso:
> from convenience point of view, should user be able to select chips and e.g copy in clipboard the content of the input?

That’s absolutely right, good catch.

Another feature we need to add at some point is editing a chip, but there are a lot of things to consider if a user is editing a chip in the middle of other chips, so it’s something we can look into on a v2.

andreapiras:
IE11 issue 
image
Android Issue 
image

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+    if (allowNew && query.trim()) {
+      this.selectItem({ label: query });
+    }
+    onBlur();
+  };
+
+  onFocus = () => {
+    this.showMenu();
+
+    this.props.onFocus();
```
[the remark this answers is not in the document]
andreapiras: I think this can be written also as this.props.onFocus && this.props.onFocus()

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+  };
+
+  onOptionSelected = (event, item) => {
+    event.preventDefault();
+    this.selectItem(item);
+  };
+
+  onInputChange = event => {
```
andreapiras: Is this not a prop?

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+      this.handleSearchDebounced();
+    }
+
+    this.setState({ query: event.target.value });
+
+    if (!multiple && selected.length > 0) {
+      this.updateSelectedValue([]);
+    }
+
+    onInputChange();
```
andreapiras: This can be null

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+
+  onKeyDown = event => {
+    const { showSuggestions, allowNew, multiple, chipSeparators, options } = this.props;
+    const { keyboardFocusedOptionIndex, query, selected } = this.state;
+    const chipsMode = !showSuggestions && allowNew && multiple;
+
+    if (chipsMode && ['Enter', ...chipSeparators].indexOf(event.key) !== -1 && query.trim()) {
+      this.selectItem({ label: query });
+
+      event.preventDefault();
```
andreapiras: there something that triggers me when I see that event.preventDefault is not at the top of the block. :)

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+    ) {
+      optionsToRender.push({
+        label: query,
+      });
+    }
+    return (
+      <div className={classNames('dropdown', { open: dropdownOpen })} id={`menu-${id}`}>
+        <ul className="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
+          {optionsToRender.map((option, idx) => {
+            return (
```
andreapiras: return is redundant here

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+                multiple,
+                optionsShown,
+                placeholder,
+                selected,
+                query,
+                maxHeight,
+              }}
+              typeaheadId={id}
+              renderChip={this.renderChip}
+              onChange={this.onInputChange}
```
andreapiras: this.onChange

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+        })}
+        onClick={this.stopPropagation}
+      >
+        <div
+          className={classNames('form-group', { 'has-error': hasError, 'has-warning': hasWarning })}
+        >
+          <div className={`input-group input-group-${size}`}>
+            {addon && <span className="input-group-addon">{addon}</span>}
+
+            <Input
```
andreapiras: I'd call this typeaheadInput  just to avoid possible collisions if we'll implement an Input component

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+              {...{
+                autofocus,
+                multiple,
+                optionsShown,
+                placeholder,
+                selected,
+                query,
+                maxHeight,
+              }}
+              typeaheadId={id}
```
andreapiras: this can be id

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+    searchDelay: SEARCH_DELAY,
+    minQueryLength: MIN_QUERY_LENGTH,
+    addon: null,
+    placeholder: null,
+    alert: null,
+    footer: null,
+    onSearch: () => {},
+    onBlur: () => {},
+    onInputChange: () => {},
+    onFocus: () => {},
```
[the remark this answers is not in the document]
gittymcgit: then we going to have to check it every time before calling. I will define noop and replace it

src/typeahead/input/Input.js
```diff
@@ -0,0 +1,167 @@
...
+
+const MIN_WIDTH = 10;
+
+export default class Input extends Component {
+  state = {
+    inputWidth: MIN_WIDTH,
+  };
+
+  static propTypes = {
+    typeaheadId: Types.string.isRequired,
```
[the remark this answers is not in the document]
gittymcgit: I need to have typeaheadId and typeaheadMenuId for aria-controls attribute. Will rename it then

src/typeahead/input/Input.js
```diff
@@ -0,0 +1,167 @@
...
+            this.inputRef.focus();
+          }}
+          style={maxHeight && { maxHeight }}
+        >
+          <div className="typeahead__input-wrapper">
+            {selected && selected.map((chip, idx) => renderChip(chip, idx))}
+
+            <input
+              type="text"
+              ref={ref => {
```
andreapiras: https://reactjs.org/docs/refs-and-the-dom.html
```class MyComponent extends React.Component {
 this.myRef = React.createRef();
  render() {
    return <div ref={this.myRef} />;
  }
}```

File: src/typeahead/input/Input.js [hunk 44]
```diff
@@ -0,0 +1,167 @@
...
+          className="form-control typeahead__input-container"
+          onClick={() => {
+            this.inputRef.focus();
+          }}
+          style={maxHeight && { maxHeight }}
+        >
+          <div className="typeahead__input-wrapper">
+            {selected && selected.map((chip, idx) => renderChip(chip, idx))}
+
+            <input
```
andreapiras: this two input have lot in common teh only diff is ``` <input
              className="typeahead__input form-control"
              id={`input-${typeaheadId}`}
              placeholder={hasValue ? '' : placeholder}
              style={hasValue ? { width: inputWidth } : {}}
            />```
So maybe you can create a getInputField(){}
And render conditionally the input

src/typeahead/input/Input.js
```diff
@@ -0,0 +1,167 @@
...
+            }}
+            className="sizer form-control typeahead__input"
+          >
+            {query}
+          </div>
+        </div>
+      );
+    } else
+      return (
+        <input
```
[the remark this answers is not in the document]
andreapiras: Sorry I addressed the comment above

src/typeahead/input/Input.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    component.setProps({
+      multiple: true,
+      selected,
+      renderChip: jest.fn(({ label }) => <span>{label}</span>),
+    });
+
+    expect(inputWrapper().find('span')).toHaveLength(selected.length);
+  });
+
+  it('passes all callbacks to input', () => {
```
andreapiras: Maybe it would be worth it to actually simulate an onfocus/onblur and check that the function prop has been called with right values ?

src/typeahead/input/Input.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(component.children()).toHaveLength(0);
+    expect(component.is('input')).toBe(true);
+  });
+
+  it('renders chips when provided ', () => {
+    const selected = [{ label: 'test1' }, { label: 'test2' }];
+    component.setProps({
+      multiple: true,
+      selected,
+      renderChip: jest.fn(({ label }) => <span>{label}</span>),
```
andreapiras: noice

src/typeahead/option/Option.js
```diff
@@ -0,0 +1,63 @@
...
+          <strong>{value.substring(highlightStart, highlightEnd)}</strong>
+          {value.substring(highlightEnd)}
+        </Fragment>
+      );
+    }
+  }
+  return value;
+};
+
+const Option = props => {
```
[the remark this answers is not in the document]
gittymcgit: I'm highlighting text inside the option. I can rename it in typeahead option and later, if we would like to implement the same functionality in select, we can merge two components

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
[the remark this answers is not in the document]
andreapiras: Naughty naughty

src/typeahead/Typeahead.js
gittymcgit: It's specifically typeahead id for aria purposes

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,445 @@
...
+    onInputChange: Types.func,
+    onFocus: Types.func,
+
+    chipSeparators: Types.arrayOf(Types.string),
+    size: Types.oneOf([Sizes.MEDIUM, Sizes.LARGE]),
+  };
+
+  static defaultProps = {
+    allowNew: false,
+    autofocus: false,
```
andreapiras: CamelCase?

src/typeahead/input/Input.js
gittymcgit: is there a benefit of actually replacing it?

andreapiras:
I don't know if this is the intended behaviour but I've noticed that when typing the options don't get filtered

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+        },
+      ],
+    };
+
+    component = shallow(<Typeahead {...props} />);
+  });
+
+  it('renders addon', () => {
+    component.setProps({ addon: <div id="test-addon" /> });
+    expect(component.find('#test-addon')).toHaveLength(1);
```
andreapiras: Please test length === 0 before

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+  });
+
+  it('renders clear button', () => {
+    component.setProps({
+      clearable: true,
+    });
+    component.instance().setState({
```
andreapiras: Normally the test shouldn't depend on the internal implementation as we may change in the future keeping the same input -> output

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      selected: [{ label: 'test' }],
+    });
+    expect(closeButton()).toHaveLength(1);
+  });
+
+  it('splits pasted value correctly', () => {
+    const chips = ['test', 'test2', 'test3'];
+    const event = {
+      preventDefault: jest.fn(),
```
andreapiras: noice

andreapiras:
image

andreapiras:
List of items is not navigable with keyboard arrows

src/typeahead/input/Input.js
andreapiras: tbh I m not sure but this is the new API and I rather use this but up to you :)

src/typeahead/typeaheadInput/TypeaheadInput.js
```diff
@@ -0,0 +1,148 @@
+/* eslint-disable jsx-a11y/no-autofocus */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+import React, { Component } from 'react';
+import Types from 'prop-types';
+import classnames from 'classnames';
+
+const DEFAULT_INPUT_MIN_WIDTH = 10;
+const noop = () => {};
```
andreapiras: meh

src/typeahead/typeaheadInput/TypeaheadInput.spec.js
```diff
@@ -0,0 +1,59 @@
...
+      onFocus: jest.fn(),
+      onPaste: jest.fn(),
+    };
+    component.setProps(extraProps);
+    input().simulate('blur');
+    input().simulate('focus');
+    input().simulate('paste');
+    input().simulate('keyDown');
+    const inputProps = input().props();
+    expect(inputProps.onBlur).toBeCalled();
```
andreapiras: Nice but Should this be toBeCalledWith() so we also have an idea of what it has been passed back?

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,464 @@
...
+  onBlur = () => {
+    const { allowNew, onBlur } = this.props;
+    const { query } = this.state;
+    this.setState({
+      isFocused: false,
+    });
+    if (allowNew && query.trim()) {
+      this.selectItem({ label: query });
+    }
+    onBlur();
```
andreapiras: should we at least pass back the event here?
and wrap in if?

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,464 @@
...
+  };
+
+  onPaste = event => {
+    const { allowNew, multiple, chipSeparators } = this.props;
+    const { selected } = this.state;
+
+    if (allowNew && multiple && chipSeparators.length > 0) {
+      event.preventDefault();
+      const value = event.clipboardData.getData('text');
+      const regex = new RegExp(chipSeparators.join('|'));
```
andreapiras: this should go inside the if as if the value is not defined there s no reason to assign regex

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,464 @@
...
+    });
+    document.removeEventListener('click', this.handleDocumentClick, false);
+  };
+
+  updateSelectedValue = selected => {
+    const { onChange, validateChip } = this.props;
+
+    const errorState = selected.some(chip => !validateChip(chip));
+    this.setState({ selected, errorState });
+    onChange(selected);
```
andreapiras: should this be called as callback of setState so we are sure that when the value gets passed outside the state value is actually being set?


```this . setState(
 { selected, errorState },
  onChange(selected);
);```

File: src/typeahead/Typeahead.js [hunk 59]
```diff
@@ -0,0 +1,464 @@
...
+
+    if (!optionsShown) {
+      this.showMenu();
+    }
+
+    if (query.length >= minQueryLength) {
+      this.handleSearchDebounced();
+    }
+
+    this.setState({ query: event.target.value });
```
andreapiras: maybe the onChange prop function should be called as callback of setState so we ll be sure that the content outside and the internal value are the same

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,464 @@
...
+
+    document.addEventListener('click', this.handleDocumentClick, false);
+  };
+
+  hideMenu = () => {
+    this.setState({
+      optionsShown: false,
+      keyboardFocusedOptionIndex: null,
+    });
+    document.removeEventListener('click', this.handleDocumentClick, false);
```
andreapiras: I think as a good practice this should all be callback of setState (even if in this case it's really rare the possible of error) as the listener should be removed once effectively the optionsShown is false

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      },
+    };
+
+    component.setProps({
+      allowNew: true,
+      multiple: true,
+      chipSeparators: [',', ' '],
+    });
+
+    component.instance().onPaste(event);
```
andreapiras: Can we not simulate the event on the element here?

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+
+    expect(
```
andreapiras: I'd try to not test the internal implementation

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
+import React from 'react';
```
andreapiras: I think the whole file should be revisited and we should removed tests on state to rather test which values are we expecting on output  so given inputA I expect component to have outputA ect

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      query,
+    });
+
+    closeButton().simulate('click', fakeEvent());
+    expect(component.state('selected')).toHaveLength(0);
+    expect(component.state('query')).toHaveLength(0);
+  });
+
+  it('moves selected items when down and up keys pressed', () => {
+    doTimes(3, () => component.instance().onKeyDown(fakeKeyDownEventForKey(KeyCodes.DOWN)));
```
andreapiras: noice

src/typeahead/typeaheadInput/TypeaheadInput.js
```diff
@@ -0,0 +1,148 @@
...
+  static defaultProps = {
+    autoFocus: false,
+    maxHeight: null,
+    placeholder: '',
+    optionsShown: false,
+    selected: [],
+
+    renderChip: noop,
+    onKeyDown: noop,
+    onBlur: noop,
```
andreapiras: rather null than empty function

src/typeahead/typeaheadOption/TypeaheadOption.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+import React, { Fragment } from 'react';
+import Types from 'prop-types';
+import classNames from 'classnames';
+
+import './TypeaheadOption.less';
+
+const hightlight = (value, query) => {
```
andreapiras: Should this function be moved into a local util folder with its own tests?

gittymcgit:
I changed tests a bit, after a discussion that we've had about it. Also fixed IE11 and android styling!

gittymcgit:
image

did that on android. Don't want to mess with scroll position

src/typeahead/Typeahead.spec.js
```diff
@@ -0,0 +1,294 @@
...
+
+    input().simulate('paste', event);
+    const renderedChips = chip().map(chipNode => chipNode.text());
+
+    expect(renderedChips.every((label, idx) => chips[idx] === label)).toBe(true);
+    expect(renderedChips).toHaveLength(chips.length);
+    expect(event.preventDefault).toBeCalled();
+  });
+
+  it('removes last selected value when backspace clicked on empty input', () => {
```
andreapiras: Much better

gittymcgit: Thanks to you!

src/typeahead/Typeahead.js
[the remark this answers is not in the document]
gittymcgit: The only thing we can do about them atm is to move them to lines they are for. I'm not sure if we can actually avoid those warnings sadly

src/typeahead/typeaheadOption/TypeaheadOption.js
gittymcgit: moved

src/typeahead/typeaheadInput/TypeaheadInput.js
gittymcgit: replaced everywhere

src/typeahead/Typeahead.spec.js
gittymcgit: changed all the tests

src/typeahead/Typeahead.js
gittymcgit: done

src/typeahead/Typeahead.js
```diff
@@ -0,0 +1,494 @@
...
+        }
+      });
+    }
+  }
+
+  componentWillUnmount() {
+    this.handleSearchDebounced.cancel();
+  }
+
+  onBlur = () => {};
```
andreapiras: Can we please remove if we don't need it ?

PR #180, opened 2019-07-10
[NP-XXX] Remove padding from top of modal body if title is not provided

rs6g10:
When we don't have a title provided; we remove the padding top from modal-body to make the modal look more appropriate.

**How did it look before?**
Screenshot 2019-07-10 at 14 28 05

How will it look now?
Screenshot 2019-07-10 at 14 27 46


src/modal/Modal.less
```diff
@@ -15,6 +15,17 @@
   }
 }
 
+.modal-body--no-title {
```
gusso: @domhorvathtw Can we get your input on this?

domhorvathtw: LGTM 🚀 
Makes sense, I understand the intent to reduce the padding in the top when there is no title present.

PR #182, opened 2019-07-17
Flexible checkbox content

fergusjordantw:
Passing through a component was impossible because of the string concatenation. Required asterisk is now just also a text node on its own which fixes this.


src/checkbox/Checkbox.spec.js
```diff
@@ -8,6 +8,8 @@ describe('Checkbox', () => {
   let component;
   let props;
 
+  const Message = ({ children }) => <Fragment>{children}</Fragment>; // eslint-disable-line react/prop-types
```
andreapiras: Proptypes can still be defined for stateless component as Message.propTypes

fergusjordantw:
@gusso the main reason I thought we should be doing this is not actually related to the input nested in the checkbox example on that `rate-alerts` project. It's more just that if we try and pass in anything that's slightly dynamic it doesn't work well. i.e. the way we use a component for translations in the calculator (pseudo example):

image

It just evaluates to `[object Object]` because it's trying to concatenate that with the required asterisk:

image

Would also happen if there was a use case where some of the text needed to be bold, etc.

andreapiras:
> @gusso the main reason I thought we should be doing this is not actually related to the input nested in the checkbox example on that `rate-alerts` project. It's more just that if we try and pass in anything that's slightly dynamic it doesn't work well. i.e. the way we use a component for translations in the calculator (pseudo example):
> 
> image
> 
> It just evaluates to `[object Object]` because it's trying to concatenate that with the required asterisk:
> 
> image
> 
> Would also happen if there was a use case where some of the text needed to be bold, etc.

But if message change shouldn't the component be rerender with the new label prop?

fergusjordantw:
> But if message change shouldn't the component be rerender with the new label prop?

@andreapiras the `retranslate` [package](https://github.com/Tankenstein/retranslate/blob/master/src/message/Message.js#L18) doesn't always just return a translated string though, at the very least it can sometimes return spans.

I'm also just looking at it from a more general thought that it should be able to handle having bold text passed into it if necessary, etc.

PR #183, opened 2019-07-21
chip component

gittymcgit:


src/chip/Chip.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    expect(component.childAt(0).text()).toEqual(props.label);
+  });
+
+  it('calls onRemove when remove button clicked', () => {
+    removeButton().simulate('click', fakeEvent());
+    expect(props.onRemove).toBeCalled();
+  });
+
+  it('renders invalid chip correctly', () => {
+    component.setProps({
```
andreapiras: I'd also check expect(component.hasClass('has-error')).toBe(false); before changing the prop

src/chip/Chip.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      valid: false,
+    });
+    expect(component.hasClass('has-error')).toBe(true);
+  });
+
+  it('render with correct className', () => {
+    component.setProps({
```
andreapiras: same as above

src/chip/Chip.less
```diff
@@ -0,0 +1,48 @@
+@import '~bootstrap/less/variables';
```
andreapiras: Do we still need this file?

PR #186, opened 2019-07-23
feat: Can show months in short format in DateInput

jacklesliewise:
Expose a new optional string prop called `monthFormat` which defaults to `long` (this is the default for `format` anyway in `getMonthNames`). If `short` is passed then will return shortened version of months (i.e `January` -> `Jan`)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/components",
-  "version": "10.1.0",
+  "version": "10.2.0",
```
taaviaasver: Could you also run `npm install` and commit the updated `package-lock` as well?

jacklesliewise: Done!

PR #187, opened 2019-07-24
Propagate states

kaada:
Propagating available states for button and process indicator.
Can now set props by doing e.g. `size={ProcessIndicator.Size.Small}`.


src/button/index.js
```diff
@@ -1 +1,3 @@
-export { default } from './Button';
+import Button from './Button';
```
oliverviljamaa: We can now keep this file as it was before 😊

src/processIndicator/ProcessIndicator.docs.js
```diff
@@ -15,21 +14,21 @@ const KNOBS = {
       type: 'select',
       label: 'Status',
       state: 'status',
-      options: processIndicatorStatuses.map(value => ({
+      options: ProcessIndicator.Status.map(value => ({
```
oliverviljamaa:
```suggestion
      options: Object.values(ProcessIndicator.Status).map(value => ({
```

src/processIndicator/ProcessIndicator.docs.js
```diff
@@ -15,21 +14,21 @@ const KNOBS = {
...
       })),
-      defaultState: { value: processIndicatorStatuses[0], label: processIndicatorStatuses[0] },
+      defaultState: { value: ProcessIndicator.Status.Processing, label: ProcessIndicator.Status.Processing },
     },
     {
       type: 'select',
       label: 'Size',
       state: 'size',
-      options: processIndicatorSizes.map(value => ({
+      options: ProcessIndicator.Size.map(value => ({
```
oliverviljamaa:
```suggestion
      options: Object.values(ProcessIndicator.Size).map(value => ({
```

PR #188, opened 2019-07-24
Flex and Box component

andreapiras:
### JIRA Ticket 
https://transferwise.atlassian.net/browse/NP-88
### Description
As part of the layout creation I though it would be useful to have a flex utility to push consistency across the way we create layout and to reduce the amount of css generated.
The two component develop (Flex and Box) allow to define a layout specifying the direction of flex and the width of the boxes per breakpoint:
Example props:
**Flex direction**  { xs: 'column-reverse', sm: 'row', md: 'column', lg: 'row' ,xl:'row-reverse'}
**Box width**  { xs: 0 sm: 200 md:1/2 lg: 1/3,xl:200}
### Demo
https://transferwise.github.io/components/branch/NP-88-grid-component/index.html


docs/utils/generateCodeBlock/generateCodeBlock.js
```diff
@@ -53,7 +53,7 @@ const generateFormattedValue = value => {
         return `{"${value || ''}"}`;
       }
 
-      return JSON.stringify(value, null, '...');
+      return JSON.stringify(value, null, '    ');
```
skywritergr: Why are you adding space here at all?

src/box/Box.docs.js
```diff
@@ -0,0 +1,191 @@
...
+import { generateCodeBlock, generateInput, generateState } from '../../docs/utils';
+import Box from './Box';
+
+import { JustifyContent, AlignItems } from '../common';
+
+const JUSTIFY_OPTIONS = JustifyContent.map(s => ({ value: s, label: s }));
+const ALIGNMENTS_OPTIONS = AlignItems.map(s => ({ value: s, label: s }));
+const WIDTH = ['0', '1 / 4', '1 / 3', '1 / 2', '1', '100', '200'];
+
+const KNOBS = {
```
skywritergr: Maybe knobs is not a great name? Is this buttons? what is it?

src/box/Box.docs.js
```diff
@@ -0,0 +1,191 @@
...
+              <h2>Box</h2>
+              <p>I&apos;m still, I&apos;m still Jenny from the Box!</p>
+              {generateCodeBlock('Box', KNOBS, this, {
+                width: { xs, sm, md, lg, xl },
+              })}
+              <p>
+                Box component is a simple html container that{' '}
+                <strong>works inside display:flex parent</strong>
+                <br />
+                The most important functionality of Box is the ability of changing its width based
```
twrdjpalmer: If I remember from discussions, the `<Box />` component is intended to be a low-level component that we're not going to use directly on a regular basis. If this is the case, it would be great to call that out in the documentation, maybe with an example of where you perceive it being used

docs/utils/generateCodeBlock/generateCodeBlock.js
andreapiras: This keeps the indentation when generating the documentation for objects which is the gray part next to the component

src/box/Box.js
```diff
@@ -0,0 +1,104 @@
...
+    </Tag>
+  ) : null;
+};
+
+Box.propTypes = {
+  alignItems: Types.oneOf([...Object.values(AlignItems)]),
+  children: Types.oneOfType([Types.arrayOf(Types.node), Types.node]),
+  extraClassNames: Types.arrayOf(Types.string),
+  justifyContent: Types.oneOf([...Object.values(JustifyContent)]),
+  mx: Types.number,
```
twrdjpalmer: Took me a little while to grok what this meant, any reason to use shorthand and not the full string `marginX`?

twrdjpalmer: Also curious as to why we only allow margin/padding in for x and y rather than each side independently

src/box/Box.docs.js
andreapiras: Knobs are the fields used in the docs to play with the props of the component. I used the same name from storybook. https://www.npmjs.com/package/@storybook/addon-knobs. I wouldn't worry to much about that as those are going to be migrated soon into the docs [repo](https://transferwise.github.io/neptune-docs/)

src/common/flexProperties.js
```diff
@@ -0,0 +1,3 @@
+export const JustifyContent = ['start', 'end', 'center', 'between', 'around'];
```
twrdjpalmer: Maybe not for discussion on this PR, but how come we separate the allow prop values from the component that is the primary consumer? 

Wouldn't it also make sense to have these as an enum so we can use them for consumption? e.g.

```jsx
<Box justifyContent={JustifyContent.center} />
```

src/flex/Flex.js
```diff
@@ -0,0 +1,70 @@
...
+  const { direction, children, mx, px, my, py } = props;
+
+  const getFlexDirection = breakpoint => {
+    return direction && direction[breakpoint] && flexDirection.indexOf(direction[breakpoint]) > -1
+      ? direction[breakpoint]
+      : 'row';
+  };
+
+  const style = {
+    [mq[0]]: {
```
twrdjpalmer: The different between `mq` vs `mx` and `my` is quite confusing.

Also have we thought about how we could make the component flexible enough to live in a post-bootstrap world should we move away? We could make mediaQueries a prop on the component and set the default to what `mq` is here

src/box/Box.js
andreapiras: Those classes were used in bootstrap in fact they mapped to m-x-number but it can be refactored to be marginX. To be honest the margin should be passed from flex and apply evenly to all children you can see that in action in the flex demo. Basically all the "columns"/box will be separated evenly

twrdjpalmer: Is equal margins something we want to enforce? It makes sense from a traditional grid point of view, but I've definitely used flex in places where I've only added margin to left or right as an example

andreapiras: It feels this kind of margin belongs more to the "content" so I'd rather apply that to the content itself and leave the flex "clean"

andreapiras: Updated the mx -> to marginX same for other short names

src/common/flexProperties.js
andreapiras: I separated them because they are shared between flex and box and may b eshared in other components too. I'd prefer to leave them as array as that makes easier to use them for props definitions

andreapiras:
> I'd be wary of introducting emotion for a single component without discussion from the guild side.

Hey, I've already asked in the guild but didn't get much responses. Some people are against some people are ok with it. Tbh the usage of it in the component is quite limited to width and direction so maybe it would also be a good case to test the solution and css in js?

src/common/flexProperties.js
Tankenstein: I'm with rich here. You can just call `Object.values` for proptypes.

PR #194, opened 2019-08-02
New theme prop for FlowNavigation

jacklesliewise:
I've exposed a new prop for FlowNavigation to change the theme - I've added a common themes file as well. At the moment the only change in FlowNavigation depending on the theme is the logo.


gusso:
Dope ✨

src/flowNavigation/FlowNavigation.js
```diff
@@ -30,7 +44,7 @@ const FlowNavigation = ({ steps, activeStep, avatarUrl, done, profileType, onClo
             />
           )}
         </div>
-        {done ? (
+        {done || theme === Themes.DARK ? (
```
anttio: Should this hiding of the stepper use something like `hideStepper` instead of it being hidden based on the theme? I mean that the functionality of a component probably shouldn't be defined which theme it is using.

src/flowNavigation/FlowNavigation.docs.js
```diff
@@ -135,17 +137,20 @@ export default class FlowNavigationDocs extends Component {
...
+            steps={this.state.selectedPreset.steps}
+            activeStep={this.state.activeStep}
+            // eslint-disable-next-line no-alert
+            onClose={this.state.onClose ? () => alert('Close clicked') : undefined}
+            avatarUrl={this.state.avatarUrl}
+            done={this.state.done}
+            profileType={this.state.profileType ? this.state.profileType.value : undefined}
+            onGoBack={this.canGoBack() ? () => this.goBack() : undefined}
+            showCloseButton={this.state.showCloseButton}
+            theme={this.state.theme ? this.state.theme.value : undefined}
```
andreapiras: I'd pass null here if not defined

src/flowNavigation/FlowNavigation.docs.js
```diff
@@ -159,6 +164,7 @@ export default class FlowNavigationDocs extends Component {
   activeStep={${this.state.activeStep}}
   steps={${this.getStringifiedSteps()}}
   done={${this.state.done}}
+  theme={${this.state.theme ? `"${this.state.theme.value}"` : undefined}}
```
andreapiras: same here

src/flowNavigation/FlowNavigation.docs.js
```diff
@@ -244,6 +250,17 @@ export default class FlowNavigationDocs extends Component {
                 onChange={this.createStateLink('onClose')}
                 checked={this.state.onClose}
               />
+              <div className="m-t-3" />
+              <label htmlFor="flow-navigation-theme">Theme</label>
+              <Select
+                id="flow-navigation-theme"
+                options={['Light', 'Dark'].map(label => ({
```
andreapiras: Can't we use Object.keys(Theme) here

src/flowNavigation/FlowNavigation.js
jacklesliewise: I agree but we don't have designs for dark theme stepper and can't show it, so at the moment the theme is dictating the functionality for a reason - although I think this should only be a temporary solution as when we do have dark theme designs we should use something like you said

src/flowNavigation/FlowNavigation.docs.js
jacklesliewise: Yeah I just copypasta'd the line for profileType:

`profileType={this.state.profileType ? this.state.profileType.value : undefined}`

Should they all be changed?

src/flowNavigation/FlowNavigation.docs.js
jacklesliewise: Sure thing!

jacklesliewise: Actually I also copypasta'd this from the profileType one:

```
options={['Personal', 'Business'].map(label => ({
                  label,
                  value: label.toUpperCase(),
                }))}
```
The reason being the labels having capital letters at the front

src/flowNavigation/FlowNavigation.js
```diff
@@ -5,18 +5,32 @@ import './FlowNavigation.less';
...
+  onGoBack,
+  theme,
+}) => (
   <div className={`tw-flow-navigation ${done ? 'tw-flow-navigation--done' : ''}`}>
     <div className="container">
       <div className="row p-t-3 tw-flow-navigation__wrapper">
         <div className="col-lg-2 col-xs-6 m-lg-t-1">
-          <div className="logo logo-primary logo-3 hidden-xs" />
+          <div
+            className={`logo ${
```
andreapiras: Please use classnames here ex: https://github.com/transferwise/components/blob/master/src/drawer/Drawer.js#L28

src/flowNavigation/FlowNavigation.js
```diff
@@ -66,6 +81,7 @@ FlowNavigation.propTypes = {
   profileType: Types.oneOf(Object.keys(ProfileType)),
   onGoBack: Types.func,
   onClose: Types.func,
+  theme: Types.string,
```
andreapiras: This one should be oneOf position: Types.oneOf(Object.keys(Themes))

src/flowNavigation/FlowNavigation.spec.js
```diff
@@ -123,4 +128,14 @@ describe('Flow navigation', () => {
     component.setProps({ onClose: undefined });
     expect(closeButton().exists()).toBe(false);
   });
+
+  it('inverts the logo if theme is dark', () => {
+    component.setProps({ theme: Themes.DARK });
```
andreapiras: Please add expect(logoInversed()).toBe(false ) before setting the props

src/flowNavigation/FlowNavigation.spec.js
```diff
@@ -123,4 +128,14 @@ describe('Flow navigation', () => {
...
+    expect(logoInversed()).toBe(true);
+  });
+
+  it('hides the stepper if theme is dark', () => {
+    component.setProps({ theme: Themes.DARK });
```
andreapiras: Please add expect(stepper().length).toBe(1);

src/flowNavigation/FlowNavigation.js
gusso: Totally fair, we went this route because we didn't have a designed version of the stepper for dark background. We also didn't want to make it available here because there's no use case yet and we don't want to encourage it.

src/flowNavigation/FlowNavigation.docs.js
andreapiras: I'd change them all tbh

src/flowNavigation/FlowNavigation.docs.js
andreapiras: But doesn't Themes have capital keys too? actually this can be rewritten as : 
```options={Object.keys(Themes).map(value => ({
                  Themes[value],
                  value,
                }))};
```

andreapiras:
Please update branch and update also changelog and package.json as well as update package-json.lock

src/flowNavigation/FlowNavigation.js
andreapiras: I'd at least add a comment about the above on line 47 with a TODO

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v10.5.0
+## Add basic dark theme option to FlowNavigation
+Currently inverts logo and hides stepper
```
oliverviljamaa: Is it expectable behaviour to hide the stepper with a theme change?

src/common/themes.js
```diff
@@ -0,0 +1,4 @@
+export const Themes = {
```
oliverviljamaa: Let's follow the [enum naming convention from Java](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) and keep it singular.

CHANGELOG.md
jacklesliewise: We had a discussion about it here https://github.com/transferwise/components/pull/194/files#r310591784

src/flowNavigation/FlowNavigation.js
```diff
@@ -1,36 +1,70 @@
 import React from 'react';
 import Types from 'prop-types';
+import classNames from 'classnames';
 
 import './FlowNavigation.less';
 import BackButton from './backButton';
 import Stepper from '../stepper';
 import Avatar from './avatar';
+import { Themes } from '../common';
```
oliverviljamaa: Are we exposing the enum to consumers somehow? If not, let's look into exporting it either from the root, or from `FlowNavigation` 😊

src/flowNavigation/FlowNavigation.js
```diff
@@ -1,36 +1,70 @@
...
+  activeStep,
+  avatarUrl,
+  done,
+  profileType,
+  onClose,
+  onGoBack,
+  theme,
+}) => (
+  <div
+    className={classNames('tw-flow-navigation', {
```
oliverviljamaa: 👏

src/common/themes.js
jacklesliewise: Will do! I based it off of `common/sizes` which has the plurality

src/flowNavigation/FlowNavigation.js
```diff
@@ -1,36 +1,70 @@
...
+          <div
+            className={classNames('logo', 'logo-3', 'hidden-xs', {
+              'logo-primary': theme === Themes.LIGHT,
+              'logo-inverse': theme === Themes.DARK,
+            })}
+          />
           <BackButton steps={steps} activeStep={activeStep} onGoBack={onGoBack} />
-          <div className={`flag flag-info logo-3 visible-xs ${onGoBack ? 'flag--hidden' : ''}`} />
+          <div
+            className={classNames('flag', 'flag-info', 'logo-3', 'visible-xs', {
```
oliverviljamaa: 👏

src/flowNavigation/FlowNavigation.js
jacklesliewise: Not at the moment - I didn't want to commit to it in case you wanted to go a different route for Theme handling in the future - happy to add it though

src/common/themes.js
oliverviljamaa: Thanks! Yeah, we've been slightly inconsistent here in the past 😢

src/flowNavigation/FlowNavigation.js
oliverviljamaa: We seem to have an `inverse` prop for a number of components, but that has never been too self-explanatory, as that's sometimes light and sometimes dark...
What other routes do you potentially see us going with in the future?

jacklesliewise: I think @gusso might have an answer for that?

PR #195, opened 2019-08-05
Fix cannot read placeholder of undefined

fergusjordantw:
@ksunandp and I looked at debugging https://rollbar.com/TransferWise/transferflow/items/296265/ for swarm day

We found the issue was happening when you typed something into the select component (in the transferFlow this was part of the calculator) that wasn't an option. See below video to see exception triggering. 

[keyboardFocusedOptionIndex.mov.zip](https://github.com/transferwise/components/files/3468281/keyboardFocusedOptionIndex.mov.zip)

We noticed while inspecting the React props that the `keyboardFocusedOptionIndex` was set to 0 when it should have been null.

`handleSearchChange` in `Select` was treating `onSearchChange` as though it setting props would happen synchronously. 

```
handleSearchChange = event => {
    this.props.onSearchChange(event.target.value);

    const hasActiveOptions = !!this.props.options.length;
    this.setState({ keyboardFocusedOptionIndex: hasActiveOptions ? 0 : null });
};
```

To get around this, we've moved the setting of `keyboardFocusedOptionIndex` to `getDerivedState`


src/select/Select.js
```diff
@@ -95,6 +95,24 @@ export default class Select extends Component {
...
 
+  static getDerivedStateFromProps(props, state) {
+    const hasActiveOptions = !!props.options.length;
+
+    if (state.open && props.searchValue !== '') {
+      if (hasActiveOptions && state.keyboardFocusedOptionIndex === null) {
+        return {
+          keyboardFocusedOptionIndex: 0,
+        };
+      } else if (!hasActiveOptions && state.keyboardFocusedOptionIndex !== null && state.open) {
```
mikknarusk: do we need the `state.keyboardFocusedOptionIndex !== null && state.open` checks?
`state.open` is already checked in the parent `if`.
for `state.keyboardFocusedOptionIndex !== null`, can't we just return `null` either way?

fergusjordantw: The `state.open` snuck in there during some merge conflicts. Will remove that. 

As for having the verbose check for `state.keyboardFocusedOptionIndex !== null`, it's my understanding that with `getDerivedStateFromProps` you really don't want to return any unnecessary state updates, and should return `null` if nothing is updating. So we should only be running the blow return explicitly when the value is changing from an integer to null.

```javascript
return {
  keyboardFocusedOptionIndex: null,
};
```

mikknarusk: makes sense 👍

src/select/Select.spec.js
```diff
@@ -302,6 +307,7 @@ describe('Select', () => {
     callSearchChangeWith('hello');
 
     expect(focusedOptionIndex()).toBe(null);
+    expect(component.state('keyboardFocusedOptionIndex')).toBe(null);
```
Tankenstein: Normally, i would not inspect the state of the component in tests, but rather the effects it has on what the component renders. This way, the internal implementation can change while the contract remains the same. In this case, i'd try reproducing the bug in tests by running an async search change handler and building a test that would fail with the old implementation :)

fergusjordantw: @Tankenstein this tripped me up for a few since it does actually render properly (without the focused className) because in render it tests for both the `keyboardFocusedOptionIndex` on state and that it can find an option that matches that in the array (it cant). So the rendering happened properly, which is why the out of sync state was causing the exception. So I've updated the test to check the actual behaviour that the out of sync state was causing, testing that the onEnter handler is handled because of the state being in sync:

```javascript
it('handles hitting enter after options are filtered', () => {
    component.setState({ keyboardFocusedOptionIndex: 0 });
    openSelect();
    component.setProps({
      onSearchChange: e => {
        component.setProps({ options: [], searchValue: e })
      }
    });
    callSearchChangeWith('hello');
    expect( () => component.simulate('keyDown', fakeKeyDownEventForKey(KEY_CODES.ENTER)) ).not.toThrow();
  });
```

PR #198, opened 2019-08-06
Add locale support to Money component

twrdjpalmer:


package.json
```diff
@@ -167,4 +167,4 @@
       "git add"
     ]
   }
-}
+}
```
oliverviljamaa: 😡

src/money/Money.spec.js
```diff
@@ -1,16 +1,21 @@
...
 
 describe('Money', () => {
   it('has formatted amount and currency', () => {
-    const component = shallow(<Money amount={4.98} currency="GBP" />);
+    shallow(<Money amount={4.98} currency="GBP" />);
+    expect(formatMoney).toBeCalledWith(4.98, 'GBP', 'en');
+  });
 
-    expect(component.text()).toBe('4.98 GBP');
+  it('formats the amount based on the locale', () => {
```
oliverviljamaa: We're duplicating a lot between those two tests, I'd perhaps just add `locale` to the existing one? No strong opinion, though.

twrdjpalmer: I think it's useful to test that in both scenarios a the `formatMoney` function is called with a locale, whether it's explicitly passed in or whether it has to resort to a default

oliverviljamaa: Valid. To be specific then, perhaps change the description to reflect that it's `en` by default?

PR #199, opened 2019-08-07
Np 80 modal dimmer update

andreapiras:
### Ticket
https://transferwise.atlassian.net/browse/DS-89
### Description
In order to have consistent animation, Modal is now using Dimmer to generate the faded background which is shared with Drawer. Also a bit of refactoring has been made to align padding between modal and drawer.

### Demo 
https://transferwise.github.io/components/branch/NP-80-modal-dimmer-update/index.html


src/modal/Modal.js
```diff
@@ -1,34 +1,34 @@
...
-  </Fragment>
+  <Dimmer open={open}>
+    <ModalDialog open={open} {...dialogProps} />
+  </Dimmer>
 );
 
-Modal.propTypes = { open: Types.bool, ...dialogPropTypes };
-Modal.defaultProps = { open: false, ...dialogDefaultProps };
+Modal.propTypes = {
+  title: Types.node,
```
mikknarusk: I'm kinda of the opinion that `Modal` shouldn't define props that it doesn't use, so everything other than `open` should only be defined in `ModalDialog`. Don't have a strong opinion on that tho

src/modal/Modal.less
```diff
@@ -1,45 +1,153 @@
...
-      border-bottom: none;
-    }
-    .modal-footer {
-      border-top: none;
-    }
+@import '~@transferwise/neptune-css/src/less/spacing.less';
+@import '~@transferwise/neptune-css/src/less/variables/_sizes.less';
+@import '~@transferwise/neptune-css/src/less/variables/_colors.less';
+
+.modal-np {
```
mikknarusk: what does `-np` stand for

oliverviljamaa: no problem

src/modal/Modal.js
oliverviljamaa: `Modal` is the one consumers use, though, so isn't that where the prop type checks are most useful? If something just gets passed in two internal components, I completely agree, and wouldn't define the types.

src/modal/Modal.less
andreapiras: this is to avoid clashing and rewrites with bootstrap. This stands for neptune but it should be removed once we figure it out how to get rid of modal from bootstrap without breaking the internet

src/modal/Modal.js
mikknarusk: Oh so since this is a public-facing component so you'd keep prop validation in the outermost layer? Thats fair, but then again, when the definitions are in `ModalDialog`, the props will be checked anyway, just one component further down the tree. Is there a difference looking from the consumer side?

If there is, that's fair. 
If there isn't, I'd keep the definitions where they make sense contextually (is that a word)

andreapiras: tbh I left ModalDialog just cause I didn't want to rewrite too much but it feels a bit redundant

src/modal/Modal.docs.js
```diff
@@ -63,11 +64,16 @@ export default class ModalDocs extends Component {
   };
 
   render() {
-    const { open, size, title, className } = this.state;
+    const { open, size, title, className, shortText } = this.state;
+    const extraPropsDocs = {
+      body: shortText ? shortbodyTex : longbodyText,
+      footer: <Button label="Action" block />,
```
mikknarusk: this shows up as `footer={[object Object]}` in the code documentation btw

src/modal/ModalDialog/ModalDialog.js
```diff
@@ -1,68 +1,116 @@
...
+  static defaultProps = {
+    title: null,
+    footer: null,
+    size: 'md',
+    role: 'dialog',
+    className: '',
+  };
+
+  onEscape = event => {
+    if (event.key === 'Escape' && this.props.onClose) {
```
OFFLlNE: `.key` works that well? Thought that works only with keyCode

src/modal/ModalDialog/ModalDialog.js
```diff
@@ -1,68 +1,116 @@
...
+
+  static defaultProps = {
+    title: null,
+    footer: null,
+    size: 'md',
+    role: 'dialog',
+    className: '',
+  };
+
+  onEscape = event => {
```
OFFLlNE: Could destructure onClose

src/modal/Modal.js
```diff
@@ -1,34 +1,34 @@
...
+  body: Types.node.isRequired,
+  footer: Types.node,
+  size: Types.oneOf(['xl', 'lg', 'md', 'sm']),
+  onClose: Types.func.isRequired,
+  open: Types.bool,
+  role: Types.string,
```
StevePole: Why do we need role?

src/modal/Modal.less
```diff
@@ -1,45 +1,166 @@
-.no-divider {
-  .modal& {
-    .modal-header {
-      border-bottom: none;
-    }
-    .modal-footer {
-      border-top: none;
-    }
+@import '~@transferwise/neptune-css/src/less/spacing.less';
```
StevePole: Why are these styles here rather than in neptune?

src/modal/Modal.spec.js
```diff
@@ -1,54 +1,23 @@
...
-      footer: 'Some footer',
-      onClose: jest.fn(),
-    };
-    shallow(<ModalAppendingToBody {...props} />);
-
-    expect(createPortal).toBeCalledWith(<Modal {...props} />, document.body);
-  });
-
-  it('has backdrop that transitions in when modal opens', () => {
+  it('has Dimmer with prop open set correctly', () => {
```
StevePole: These test aren't that great, they don't cover any of the important functionality provided by these props:

when a title is supplied
 - it should display the title in the modal

when body content is supplied
 - it should display the body content in the modal

when a footer is supplied
 - it should display the footer in the modal

when a size is supplied
 - it should render the appropriate size dialog


when the close button is clicked:
 - it should close the modal
 - it should call the onClose handler

when the modal backdrop (dimmer?) is clicked
 - it should ...? (I don't actually know what happens in this case, tests would make it clear)

Existing tests could be reformatted so it's clearer what's being tested:

when the modal is set to 'open'
 - it should display a dimmer that's also set to open

We also don't test any of the new props:

StevePole: Looking at this again, it looks like it's responsibility is to pass these props down.  So 

when a title is supplied
 - it should display the title in the modal

Would become

when a title is supplied
 - it should pass it to the ModalDialog

If it doesn't own the click button, it should be

when the ModalDialog triggers onClose 
 - it should trigger Modal.onClose

src/modal/Modal.js
```diff
@@ -1,34 +1,34 @@
...
+  </Dimmer>
 );
 
-Modal.propTypes = { open: Types.bool, ...dialogPropTypes };
-Modal.defaultProps = { open: false, ...dialogDefaultProps };
+Modal.propTypes = {
+  title: Types.node,
+  body: Types.node.isRequired,
+  footer: Types.node,
+  size: Types.oneOf(['xl', 'lg', 'md', 'sm']),
```
dozortsev: please, use `Sizes` here

src/modal/Modal.js
dozortsev: even if for some reason we need this prop we should define what values it accepts via `Types.oneOf([...])`

andreapiras: It seems to be redundant so I remove it

src/modal/ModalDialog/ModalDialog.js
```diff
@@ -1,68 +1,117 @@
...
+    if (event && (event.keyCode === KEY_CODES.ESCAPE || event.key === 'Escape') && this.props.onClose) {
```
dozortsev: why here checking for `ESC` we need this ` || event.key === 'Escape'` part?

anyway if it's so necessary please create a constant object in `keyCodes.js`, or I would probably extend the existing one to have something like this:

```javascript
export default {
  ESCAPE: { keyCode: 27, key: 'Escape' },
  TAB: { keyCode: 9, key: 'Tab' },
  ...
};
```

src/modal/Modal.js
dozortsev: in both components `Modal` and `ModalDialog`)

src/modal/Modal.js
dozortsev: @andreapiras +1 to delete _ModalDialog_ component, as now after migration to _Dimmer_ it looks completely redundant for me

src/modal/Modal.less
andreapiras: At the moment bootstrap modal rely also on some js that push inline styles and switch classes so I can still put it on neptune but I can't remove the existing ones

src/modal/ModalDialog/ModalDialog.js
andreapiras: KeyCode seems to be deprecated. 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode. 
I followed this for key detection:
https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a

src/modal/Modal.spec.js
andreapiras: To be honest I'm not convinced of the utility of this extra Layer (ie Modal) I don't think ModalDialog is going to be reused as separate entity so I think maybe it's better If I remove it completely and group so we ll have only Modal

StevePole: Makes sense 👍

src/modal/Modal.js
```diff
@@ -1,34 +1,145 @@
...
+      ...otherProps
+    } = this.props;
+
+    // These should be replaced with props in breaking change.
+    const isCompact = this.checkSpecialClasses('compact');
+    const noDivider = this.checkSpecialClasses('no-divider');
+
+    return (
+      <Dimmer open={open} onClose={closeOnClick && onClose}>
+        <CSSTransition
```
StevePole: Where's this guy from?

src/modal/Modal.spec.js
```diff
@@ -1,90 +1,181 @@
...
-  it('has backdrop that transitions in when modal opens', () => {
-    component.setProps({ open: false });
-    expect(
-      backdropParent().matchesElement(
-        <WithInTransition>
-          <Backdrop />
-        </WithInTransition>,
-      ),
-    ).toBe(true);
+    it('with title', () => {
```
StevePole: No idea what this does without reading the implementation

src/modal/Modal.js
andreapiras: http://reactcommunity.org/react-transition-group/css-transition

src/modal/Modal.spec.js
andreapiras: Not sure what do you mean, but I've updated the text anyway

PR #201, opened 2019-08-09
Spinner component

jacklesliewise:
large


StevePole:
Screenshot please :-)

sergonius:
Perhaps something to consider is having a possibility to pass children to the spinner component.

Something like Ant has:
https://ant.design/components/spin/

Seems a common enough use-case and not a lot of extra work.

domhorvathtw:
Could be just me, but isn't it weird in terms of naming, that there is **no spinning motion** in the animation of this component? If it's going to be used as a loading indicator, can we rename it to something like `<Loader />` `<LoadingIndicator />`?

We already have a `<Loader />` though, that's the one with the flag. 
And we also have a `<ProcessIndicator />`, which is used for multiple states (failed, succeeded), not just loading.  

Any thoughts?

jacklesliewise:
Closing this in favour of #232

PR #204, opened 2019-08-15
Import single functions from lodash instead of whole package

oliverviljamaa:


src/typeahead/Typeahead.js
```diff
@@ -4,7 +4,8 @@
 
 import React, { Component } from 'react';
 import Types from 'prop-types';
-import { debounce, clamp } from 'lodash';
+import debounce  from 'lodash.debounce';
```
stepan-leibo:
```suggestion
import debounce from 'lodash.debounce';
```

src/typeahead/Typeahead.js
```diff
@@ -4,7 +4,8 @@
...
+import clamp  from 'lodash.clamp';
```
stepan-leibo:
```suggestion
import clamp from 'lodash.clamp';
```

PR #218, opened 2019-08-20
feat: can now disable options in Select component

jacklesliewise:
open
open_hover


StevePole:
Can you add a screenshot please?

In the angular component, which also uses bootstrap, we get the styling by applying a 'disabled' class on the `li` rather than an attribute on the anchor.  Just wondering if they look the same.

StevePole:
What happens when you use keyboard controls? Does it skip that option or highlight it?  If it highlights, is it selectable using keyboard?

src/select/Select.spec.js
```diff
@@ -244,6 +244,32 @@ describe('Select', () => {
     expectDropdownToBe().open();
   });
 
+  it('renders disabled options', () => {
```
StevePole: Could you break this test up?  You test that it exists( which is what your test describes), but you also test some other things: that the dropdown doesn't close on click, and that the onChange isn't called.  I like to use given, when, 'then' (it should) syntax, and separate my test setup from my expectations:
```
describe('given a set of options where some are disabled', () => {
  beforeEach(() => {
    component.setProps({
      options: [{ label: 'a thing' }, { label: 'a disabled thing', disabled: true }],
    });
  });

  describe('when opened', () => {
    beforeEach(() => {
      openSelect();
    });

    it('should render a disable links for any disabled options', () => {
      expect(component.find('li.tw-dropdown-item').first().text()).toEqual('a thing');
    });

    describe('when a disabled option is clicked' () => {
      beforeEach(() => {
        component.find('li.tw-dropdown-item').at(1).simulate('click', fakeEvent());
      });

      it('should not close the dropdown', () => {
        expectDropdownToBe().open();
      });

      it('should not call the onChange callback', () => {
        expect(props.onChange).not.toBeCalled();
      });
    });
  });
});
```

I would also test the keyboard functionality:
- when using the keyboard to navigation the options
  - it should skip disabled options

jacklesliewise: Sure thing!

PR #219, opened 2019-08-20
flowNavigation hover label renders jsx not strings->html

kaada:
flowNavigation hover label prefers jsx over a string being evaluated to html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-# v11.1.1
-## flowNavigation hover label renders jsx not strings->html
-flowNavigation hover label prefers jsx over a string being evaluated to html
+# v12.0.0
+## flowNavigation hover label renders jsx not strings->HTML
+flowNavigation hover label no longer accepts rendering strings->HTML. hoverHTML-flag is removed.
+Any Types.node, e.g. string or element/jsx, are accepted instead.
 
 # v11.1.0
 ## Add disabled property to options in Select component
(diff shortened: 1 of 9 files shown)


src/flowNavigation/FlowNavigation.js
```diff
@@ -91,8 +91,7 @@ FlowNavigation.propTypes = {
     Types.shape({
       label: Types.string.isRequired,
       onClick: Types.func,
-      hoverLabel: Types.string,
-      hoverHTML: Types.bool,
+      hoverLabel: Types.oneOfType([Types.element, Types.string]),
```
oliverviljamaa: I believe we should just allow `Types.node` (`string` or `element`, basically), and create a breaking (major version change) release 😊 What do you think?

kaada: Oh, yea. That's a good idea. Fixing.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v12.0.0
+## flowNavigation hover label renders jsx not strings->HTML
+flowNavigation hover label no longer accepts rendering strings->HTML. hoverHTML-flag is removed.
```
oliverviljamaa: Please check that docs are up to date as well (for example, `FlowNavigation.docs` seems to still use `hoverHTML`) 😊

kaada: :panic:

PR #220, opened 2019-08-21
Button label jsx

kaada:
Button label supports jsx, e.g. doing
`label={<Message>request-money.form.generate</Message>}`


src/button/Button.js
```diff
@@ -59,7 +59,7 @@ Button.propTypes = {
   state: Types.oneOf(Object.values(State)),
   block: Types.bool,
   onClick: Types.func.isRequired,
-  label: Types.string.isRequired,
+  label: Types.node.isRequired,
```
dozortsev: with this change you introducing some breaking changes

dozortsev: I would actually allow `string` label as well, just use `Types.oneOfType([Types.string, Types.node])...`

ProstoSanja: https://reactjs.org/docs/typechecking-with-proptypes.html

Node: anything that can be rendered: string, number, element or array.

Isn't it redundant to add node+string?

dozortsev: sorry, you're right

PR #222, opened 2019-08-22
Docs hotfix

andreapiras:
This PR fixed a bunch of bugs occurring in components. Please navigate by commit for easier CR


src/typeahead/Typeahead.js
```diff
@@ -392,6 +392,7 @@ export default class Typeahead extends Component {
         <ul className="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
           {optionsToRender.map((option, idx) => (
             <TypeaheadOption
+              key={`${option.label}${idx.toString()}`}
```
dozortsev:
```
`${idx.toString()}`
```

should be enough

src/ssr.spec.js
```diff
@@ -51,9 +51,19 @@ describe('Server side rendering', () => {
...
+    Typeahead: { size: 'md' },
+    Sticky: { position: 'top' },
+  };
+
   componentNames.forEach(componentName => {
     it(`works for ${componentName} components`, () => {
       const Component = components[componentName];
+      if (overrideProps[componentName]) {
+        const propToOverrideKey = Object.keys(overrideProps[componentName])[0];
+        allProps[propToOverrideKey] = overrideProps[componentName][propToOverrideKey];
```
dozortsev: IMHO, this code looks confusing

what exactly where you trying to fix here?

andreapiras: Some components have the same named props but allow different values. That was throwing warnings on the console. This allow you to specify different values for same named props

src/typeahead/Typeahead.js
andreapiras: Not sure about it as this would be basically the equivalent of using idx only which is not allowed

andreapiras: https://reactjs.org/docs/lists-and-keys.html

taaviaasver: [Not important at all, just a really tiny thing that I noticed 🙃]
`.toString()` is not needed, because template literals convert number to string anyway:
```
var x = 10;
`${x.toString()}` -> "10" (string)
`${x}` -> "10" (string)
```

PR #223, opened 2019-08-23
Flex box expose

andreapiras:
This PR exposes Flex and Box so they can be used by consumers. It also enhance the type check for Flex children and improve the testing passing an actual Box to Flex


src/flex/Flex.js
```diff
@@ -31,17 +31,17 @@ const Flex = props => {
     },
   };
 
-  return (
+  return children || children.length ? (
```
dozortsev: this check will pass if `children` is an empty array `!![]` equals `true`

src/flex/Flex.js
```diff
@@ -31,17 +31,17 @@ const Flex = props => {
...
     <div className="d-flex" css={style}>
       {React.Children.map(children, child => {
-        if (child) {
+        if (child && child.type && child.type.name === 'Box') {
```
dozortsev: so this component should accept only `Box` as children?

andreapiras: Nope you can pass other children but if you pass box is going to pass down the extrapros

src/flex/Flex.js
andreapiras: Actually this check is redundant

PR #226, opened 2019-08-23
Remove flex and box

andreapiras:
Remove Flex and Box from export as they should be used only internally as building blocks for layout pages. Direct use is discouraged and will easily introduce inconsistencies.


CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
+# v12.2.2
+## Remove export flex and box.
+
 # v12.2.1
 ## Fixed modal scroll.
 
 # v12.2.0
-## Exposed flex and box.
+## Update flex and box.
```
OFFLlNE: naughty

PR #235, opened 2019-08-29
support node in moneyinput note

ProstoSanja:
Support for JSX in currency object note.
Useful for working with translations

also this is supported down the line in `select`

Same reason as #220


src/moneyInput/MoneyInput.js
```diff
@@ -11,7 +11,7 @@ const Currency = Types.shape({
   value: Types.string,
   label: Types.string,
   currency: Types.string,
-  note: Types.string,
+  note: Types.node,
```
dozortsev: please update tests

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# v12.3.4
+## MoneyInput selected note supports jsx
+MoneyInput selected note supports jsx, e.g. doing
+selectedCurrency = {
+    value: targetCurrency,
+    label: targetCurrency,
+    note: (<Message>targetCurrency</Message>)
+    currency: targetCurrency,
+}
```
dozortsev: this code example feels redundant here, the why how it will works already clear from the code

ProstoSanja:
fond a big problem with this, while trying to implement, gonna close for the time

PR #236, opened 2019-08-30
feat: mode prop for DateInput

jacklesliewise:
We got a design for a `DateInput` with month and year only, we were going to do it locally but thought I could try put it into the `components` library - it's a bit weird in that it changes the date string to a `YYYY-MM`, let me know what you think

Screenshot 2019-08-30 at 14 18 51


src/dateInput/DateInput.js
```diff
@@ -25,6 +25,7 @@ class DateInput extends PureComponent {
     monthLabel: Types.string,
     yearLabel: Types.string,
     monthFormat: Types.string,
+    monthYearOnly: Types.bool,
```
dozortsev: ideally I would better introduce list of modes for this, e.g:

- `DAY_MONTH_YEAR_MODE` for three inputs (default one)
- `MONTH_YEAH_MODE` for month and year inputs

I think this is more scalable approach rather than bool prop, as tomorrow we might have a case when we need ask only for day & month

jacklesliewise: Good shout, added in 1206716

src/dateInput/DateInput.js
```diff
@@ -61,10 +68,17 @@ class DateInput extends PureComponent {
...
+  getDateAsString = date => {
+    const { mode } = this.props;
+    if (mode === MODE.MONTH_YEAR) {
+      return [date.getFullYear(), `0${date.getMonth() + 1}`.slice(-2)].join('-');
+    }
+    return [
+      date.getFullYear(),
+      `0${date.getMonth() + 1}`.slice(-2),
+      `0${date.getDate()}`.slice(-2),
+    ].join('-');
```
dozortsev: I would use `switch` statement here to even more ensure that `MODE.DAY_MONTH_YEAR` will be default one:

```javascript
switch (mode) {
  case MODE.MONTH_YEAR:
    return ...
  case MODE.DAY_MONTH_YEAR:
  default:
    return ...
}
```

src/dateInput/DateInput.view.spec.js
```diff
@@ -292,4 +292,26 @@ describe('Date Input Component', () => {
...
+      component = shallow(<DateInput {...{ ...props, ...extraProps }} />);
+    });
+
+    it('should only display month and year inputs', () => {
+      expect(component.exists(YEAR_SELECTOR)).toBe(true);
+      expect(component.exists(MONTH_SELECTOR)).toBe(true);
+      expect(component.exists(DAY_SELECTOR)).toBe(false);
+    });
+
+    it('should produce MM-YYYY date string', () => {
```
dozortsev: maybe would be better still have `DD-MM-YYYY`? in this case day always will be `00`, what day you will store in DB?

jacklesliewise: I tried that but because `00` isn't a valid date the entire date string becomes invalid and the component would wipe the day and year away from the view, whereas `MM-YYYY` is still a valid date string

src/dateInput/DateInput.js
jacklesliewise: Added in 5aae438

PR #241, opened 2019-09-03
Snackbar

sergonius:


sergonius:
Needs:
- [ ] Tests

OFFLlNE:
Needs pics

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,59 @@
...
+  font-size: 14px;
+  line-height: 20px;
+  display: inline-flex;
+  border-radius: 2px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.25);
+  display: inline-flex;
+  align-items: center;
+  justify-content: space-between;
+  text-align: left;
+  margin: 0 16px;
```
OFFLlNE: could use spacer here I believe

OFFLlNE: And some other spots below this line

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    component = mount(
+      <SnackbarProvider timeout={1000}>
+        <Container />
+      </SnackbarProvider>,
+    );
+  });
+
+  it('displays a single snack for the given duration with the given text', async () => {
+    expect(snackbar().text()).not.toContain(props.text);
+    component.find('.button-trigger').simulate('click');
```
OFFLlNE: might as well create a function for this component.find button

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    expect(snackbar().text()).toContain(props.text);
+    expect(snackbar().length).toBe(1);
+
+    await wait(timeout + CSS_TRANSITION_DURATION + 5);
+
+    expect(snackbar().text()).not.toContain(props.text);
+  });
+
+  function snackbar() {
```
OFFLlNE: lets change this into an arrow function

Tankenstein: No reason to. We can use function hoisting here.

Tankenstein:
@OFFLlNE Doesn't need pics. Every PR of components has its own build you can play with: 
image

src/snackbar/SnackbarProvider.js
```diff
@@ -0,0 +1,52 @@
...
+SnackbarProvider.propTypes = {
+  children: PropTypes.node.isRequired,
+  timeout: PropTypes.number,
+};
+
+SnackbarProvider.defaultProps = {
+  timeout: 4500,
+};
+
+export function wrapComponent(Comp) {
```
twrdjpalmer: Would be nice to rename `wrapComponent` to something more obvious—I'ma have no idea what it does when I see it in use.

Have we also thought about how we should handle multiple snackbars being triggered at the same time? Seems like something the provider should handle. Looks like we switch out the current one in favour of the newest which could lead to a sub-par experience if the former one is the what the user is imminently about to interact with

src/snackbar/SnackbarProvider.js
```diff
@@ -0,0 +1,52 @@
...
+  return function WrappedComponent(props) {
+    const { createSnackbar } = useContext(SnackbarContext);
+
+    return <Comp createSnackbar={createSnackbar} {...props} />;
```
twrdjpalmer: Why is it up to the consumer to create the snack bar? I find the API really confusing

src/snackbar/SnackbarProvider.js
fergusjordantw: @twrdjpalmer only having one snackbar visible at a time (being the most recent one that was triggered) was a decision made by the designers. 

> If an instance is triggered while another is being displayed, we animate away the one on the screen before showing the new one. This helps keep the relevance of the message being displayed. This feels more appropriate for our use cases than visually stacking or queueing them.

sergonius: The single snackbar thing is as per spec, so it was definitely thought about. You can take it up @gusso if you are curious.

As for the function name... I just took that from another snackbar project and didn't really put any thought into it, but how does `withCreateSnackbar` sound?

twrdjpalmer: Cool, glad it's been thought about! After a quick chat with Fergus about the component and it's intended uses, I think it would be great to include documentation which includes that info, e.g., when and why you should use a snackbar, when it should contain an action and what sort of actions are applicable to inclusion. Should help mitigating miss use by having them pop up left right and center or having super important actions in them 😄

twrdjpalmer: Naming wise, they're both improvements 👍

src/snackbar/SnackbarProvider.js
sergonius: Can you elaborate what you find confusing?
Currently, you just use the provider and a HOC (or if you want to, use the the exported context directly) to access the snack-creating function. I guess the docs could reflect all of that better, but I am open to suggestions as well.

src/snackbar/Snackbar.docs.js
```diff
@@ -0,0 +1,111 @@
...
+                  <SnackbarConsumer>
+                    {({ createSnackbar }) => (
+                      <Button
+                        size="block"
+                        onClick={() =>
+                          createSnackbar({
+                            text,
+                            action: withAction
+                              ? {
+                                  label: 'Bite me',
```
mikknarusk: noice

src/snackbar/Snackbar.docs.js
```diff
@@ -0,0 +1,111 @@
...
+                                  // eslint-disable-next-line no-alert
+                                  function: () => alert(`I can't believe you've done this.`),
```
mikknarusk: not sure if I like `function` as a property name. It's allowed technically but still a reserved word and thus confusing.
Also I think it's not descriptive enough, eg when will it be called? Shoud it be `onClick` (not sure as I haven't gotten far enough yet to see what the action is)

sergonius: I completely agree.

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,101 @@
...
+import CSSTransition from 'react-transition-group/CSSTransition';
+
+import './Snackbar.less';
+
+export const CSS_TRANSITION_DURATION = 400;
+
+class Snackbar extends Component {
+  state = {
+    visible: false,
+    snack: '',
```
mikknarusk: i find it weird that the state is called `snack` and the prop is `text` while they're kinda the same thing

sergonius: You are right, they are exactly the same thing 😖

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,101 @@
...
+    } else if (prevProps.timestamp !== timestamp) {
+      clearTimeout(this.timeout);
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ visible: false }, () => {
+        setTimeout(() => {
+          this.setState({ snack: text, visible: true });
+        }, CSS_TRANSITION_DURATION);
+      });
+
+      this.timeout = setTimeout(() => {
```
mikknarusk: This block of code is a bit confusing, so correct be if I'm wrong here.

If I understand correctly the `else if` will be called when one snackbar is active and another is triggered. In that case, shouldn't line 48 be **after** the `setTimeout` where we set `visible: true` so that the closing timer starts only when it's actually shown. Might be wrong tho, lots of transitions and timeouts here...

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,101 @@
...
+        </CSSTransition>
+      </div>
+    );
+  }
+}
+
+Snackbar.propTypes = {
+  action: PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    function: PropTypes.func.isRequired,
```
mikknarusk:
```suggestion
    onClick: PropTypes.func.isRequired,
```

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,60 @@
...
+    border: 0;
+    background: none;
+    outline: 0;
+    font: inherit;
+    color: @brand-light-blue;
+    font-weight: @font-weight-semi-bold;
+  }
+}
+
+[class*='tw-snackbar__text-container'] {
```
mikknarusk: I'm lazy, what does this do?

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,60 @@
...
+.tw-snackbar__text-container-enter,
+.tw-snackbar__text-container-exit {
+  transform: translateY(calc(100% + 10px));
+}
+
+.tw-snackbar__text-container-enter-active {
+  transform: translateY(~'-@{spacer}');
+
+  @media only screen and (min-width: @screen-sm-min) {
+    @desktop-y: calc(@spacer * -2.5);
```
mikknarusk: what's dis

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,60 @@
...
+  transition: 400ms cubic-bezier(0.215, 0.61, 0.355, 1);
+}
+
+.tw-snackbar__text-container-enter,
+.tw-snackbar__text-container-exit {
+  transform: translateY(calc(100% + 10px));
+}
+
+.tw-snackbar__text-container-enter-active {
+  transform: translateY(~'-@{spacer}');
```
mikknarusk: haven't seen this before, what will this evaluate to? (this is just me not knowing stuff about less)

mikknarusk:
Screenshot 2019-09-04 at 11 23 06

Something there is misaligned I think. Can't really tell which

package.json
```diff
@@ -168,4 +168,4 @@
       "git add"
     ]
   }
-}
+}
```
oliverviljamaa: 🐱

src/snackbar/Snackbar.js
fergusjordantw: @mikknarusk one problem setting that after the timeout where visible is set to true is if it's triggered again during the `CSS_TRANSITION_DURATION`, before the new timeout has been created, it won't get cleared via `clearTimeout(this.timeout);`. Probably shouldn't happen in a real use-case, but one way around that could be to set the second `setTimeout` to use `CSS_TRANSITION_DURATION + timeout`

src/snackbar/Snackbar.spec.js
OFFLlNE: Less lines

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,60 @@
...
+  text-align: center;
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  z-index: 2;
+}
+
+.tw-snackbar__text {
+  background: @brand-navy;
```
gusso: This should be `@brand-navy-minus-10`

gusso:
> Screenshot 2019-09-04 at 11 23 06
> 
> Something there is misaligned I think. Can't really tell which

The action button contains paddings all around, which causes the misalignment. It's good that it does, so the target area is comfortable to click/tap, but the baseline of all text needs to be vertically aligned.

We also need to compensate the apparent margin on the right when the action is present, the total of padding+margin on the right needs to be 16px, to mirror what happens on the left side.

gusso:
The aim is for the total height of the component with a single line to be `48px`, so apparent top and bottom paddings should be `14px`. The actual CSS value might have to be different to play well with the paddings of the button though.

gusso:
paddings

src/snackbar/SnackbarProvider.js
twrdjpalmer: I was originally confused as I was thinking that `wrapComponent` would create me a snackbar component which I can render somewhere... having it pass in `createSnackbar` had me second guessing but not really sure of what it's doing.

Thinking about it more, I think I understand now, but correct me if I'm wrong, `createSnackbar` is a function which can trigger a snackbar to appear on screen? If this is the case I don't think `createSnackbar` is very descriptive as it suggests to me that it'll return a snack bar for me to render somewhere, as if it were a factory function.

Perhaps it would be better name the function after what the end goal is? E.g., `notify` which speaks more to the outcome of the user seeing a snackbar appear on screen

src/snackbar/Snackbar.js
```diff
@@ -68,7 +68,7 @@ class Snackbar extends Component {
           unmountOnExit
         >
           <span className="tw-snackbar__text p-a-2">
-            {snack}
+            {text}
```
twrdjpalmer: What if we used `children` so it's more like a regular ole HTML element?

sergonius:
Got the padding and margin things resolved 🐱

src/snackbar/Snackbar.js
sergonius: Not much would change. The `Snackbar` component is only used by the `SnackbarProvider` itself.

src/snackbar/Snackbar.js
sergonius: Thanks for pointing it out. There were also related shenanigans that I got resolved thanks to this comment. 🥳

src/snackbar/Snackbar.js
twrdjpalmer: Ah, I just envisage a need to be able to bold certain text or w/e so maybe `node` as a type would be better, and then `text` wouldn't make too much sense 😄  all good either way though

gusso:
Found a couple more things related to mobile:

- Margins should be `16px` from the edges instead of `8px`.
- The bar is getting cropped on the rounded edges of the iPhone X/Xs/Xr,Xs Max, we need to use the `env(safe-area-inset-bottom)` [CSS function](https://webkit.org/blog/7929/designing-websites-for-iphone-x/) on the bottom distance to stop this from happening.

Screenshot 2019-09-04 at 1 52 15 pm

sergonius:
> Found a couple more things related to mobile:
> 
> * Margins should be `16px` from the edges instead of `8px`.
> * The bar is getting cropped on the rounded edges of the iPhone X/Xs/Xr,Xs Max, we need to use the `env(safe-area-inset-bottom)` [CSS function](https://webkit.org/blog/7929/designing-websites-for-iphone-x/) on the bottom distance to stop this from happening.
> 
> Screenshot 2019-09-04 at 1 52 15 pm

Gotcha,  should be resolved now!

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,109 @@
...
+
+    this.timeout = setTimeout(() => {
+      this.setState({ visible: false });
+    }, timeout);
+  };
+
+  componentDidUpdate(prevProps) {
+    const { text, timeout, timestamp, action } = this.props;
+
+    if (!prevProps.text) {
```
andreapiras: It looks to me that here we are trying to mirror state to props maybe it's worth to have a look at getDerivedStateFromProps

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,64 @@
+@import '~bootstrap/less/variables.less';
```
andreapiras: As discussed this should live in neptune-css

src/snackbar/Snackbar.less
```diff
@@ -0,0 +1,64 @@
...
+
+.tw-snackbar {
```
andreapiras: we ll CR this in Neptune-css PR but I don't think we need prefix here

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,106 @@
...
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ visible: true, text, action }, () => {
+        this.timeout = setTimeout(() => {
```
andreapiras: Can we use this.setLeaveTimeout() here?

src/snackbar/index.js
```diff
@@ -0,0 +1,3 @@
+import Snackbar from './Snackbar';
```
andreapiras: Optional: Can we use ```export { default } from './Snackbar'; ```?

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,71 @@
...
+      <SnackbarProvider timeout={1000}>
+        <Container />
+      </SnackbarProvider>,
+    );
+    snackbar = () => component.find(Snackbar);
+    buttonTrigger = () => component.find('.button-trigger');
+  });
+
+  it('displays a single snack for the given duration with the given text', async () => {
+    expect(snackbar().text()).not.toContain(props.text);
```
andreapiras: Should we maybe add? ```expect(snackbar().length).toBe(0);```

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    buttonTrigger().simulate('click');
```
andreapiras: Why do we need three clicks?

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    buttonTrigger().simulate('click');
+    buttonTrigger().simulate('click');
+    expect(snackbar().text()).toContain(props.text);
+    expect(snackbar().length).toBe(1);
+
+    jest.advanceTimersByTime(timeout + CSS_TRANSITION_DURATION + 5);
+
```
andreapiras: Maybe here again: Should we maybe add? expect(snackbar().length).toBe(0);?

src/snackbar/Snackbar.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    expect(onDidUpdate).toHaveBeenCalledTimes(1);
+
+    props.text = '';
+    buttonTrigger().simulate('click');
+    buttonTrigger().simulate('click');
+    expect(onDidUpdate).toHaveBeenCalledTimes(1);
+
+    props.text = 'doge';
+    buttonTrigger().simulate('click');
+    jest.advanceTimersByTime(CSS_TRANSITION_DURATION * 2 + 1000);
```
andreapiras: Why do we need this?  jest.advanceTimersByTime(CSS_TRANSITION_DURATION * 2 + 1000);
Component should be called immediately no?

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v12.5.2
```
Tankenstein: Minor, not patch :)

src/index.js
```diff
@@ -33,6 +33,8 @@ export { default as Drawer } from './drawer';
 export { default as Sticky } from './sticky';
 export { default as Chip } from './chip';
 export { default as Typeahead } from './typeahead';
+export { default as SnackbarProvider } from './snackbar/SnackbarProvider';
+export { SnackbarConsumer, SnackbarContext } from './snackbar/SnackbarProvider';
```
Tankenstein: Do these in one export

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,106 @@
...
+export const CSS_TRANSITION_DURATION = 400;
+
+export class Snackbar extends Component {
+  state = {
+    visible: false,
+    text: '',
+  };
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (!nextProps.text) return false;
```
Tankenstein: Can we always use squigglies?

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    return true;
+  }
+
+  setLeaveTimeout = () => {
+    const { timeout } = this.props;
+
+    this.timeout = setTimeout(() => {
+      this.setState({ visible: false });
```
Tankenstein: Can we clear this timeout in `componentWillUnmount`?

src/index.js
```diff
@@ -33,6 +33,11 @@ export { default as Drawer } from './drawer';
 export { default as Sticky } from './sticky';
 export { default as Chip } from './chip';
 export { default as Typeahead } from './typeahead';
+export {
+  default as SnackbarProvider,
+  SnackbarConsumer,
+  SnackbarContext,
```
Tankenstein: could you export the hook as well?

src/snackbar/Snackbar.js
```diff
@@ -0,0 +1,112 @@
...
+      this.setState({ visible: true, text, action }, () => {
+        this.setLeaveTimeout();
+      });
+    } else if (prevProps.timestamp !== timestamp) {
+      clearTimeout(this.timeout);
+
+      if (this.state.visible) {
+        // eslint-disable-next-line react/no-did-update-set-state
+        this.setState({ visible: false }, () => {
+          setTimeout(() => {
```
Tankenstein: Let's clear this timeout as well

PR #244, opened 2019-09-04
Tweak label prop of Checkbox component to accept node

dozortsev:


src/checkbox/Checkbox.js
```diff
@@ -27,7 +27,7 @@ Checkbox.propTypes = {
   checked: Types.bool,
   required: Types.bool,
   disabled: Types.bool,
-  label: Types.string.isRequired,
+  label: Types.node.isRequired,
```
andreapiras: this is a breaking change. Can it not accept both?

andreapiras:
Tests are failing :)

src/checkbox/Checkbox.js
dozortsev: `node` already does this, check the [doc](https://reactjs.org/docs/typechecking-with-proptypes.html) 🙂

> // Anything that can be rendered: numbers, strings, elements or an array

dozortsev:
just a little code formatting issue

PR #245, opened 2019-09-04
Refactor Avatar to not use deprecated methods

twrdjpalmer:
1. Removes dependency on deprecated componentWillReceiveProps
2. Simplifies the logic of showing/hiding the profile image

See https://github.com/transferwise/components/commit/32c180d2bc8e5c553f757d9832e76b6dcf0367e5 for the traditional refactor.
See https://github.com/transferwise/components/pull/245/commits/93d37f5468fbb82ad0078f15b67689c5c1a12390 for a hooks implementation I added after consulting with @Tankenstein 

**Todo**
- [x] Add changelog, bump version


src/flowNavigation/avatar/Avatar.js
```diff
@@ -1,71 +1,44 @@
...
+            <img
+              className="tw-avatar__image"
+              src={url}
+              alt="avatar"
+              onError={() => setImageLoadError(true)}
+            />
+          )}
+        </div>
+      ) : (
+        <div className="icon icon-bank" />
```
andreapiras: Can we use tw cion for this?
https://transferwise.github.io/icons/

src/flowNavigation/avatar/Avatar.js
```diff
@@ -1,71 +1,44 @@
...
-        {this.shouldShowPersonalAvatar() ? (
-          this.renderPersonalAvatar()
-        ) : (
-          <div className="icon icon-bank" />
-        )}
-      </div>
-    );
-  }
+export default function Avatar({ url, profileType }) {
+  const [hasImageLoadError, setImageLoadError] = useState(false);
```
andreapiras: It's happening <3

andreapiras:
LGTM  👏 . Make sure u update version number in package.json and changelog and also run npm install to update package-lock

src/flowNavigation/avatar/Avatar.less
```diff
@@ -0,0 +1,9 @@
+.tw-avatar__icon-container {
```
nabaroa: It would be awesome if this component should consume the CSS that needs from Neptune CSS https://github.com/transferwise/neptune-css and not have a .less file on it. The advantage of having it in Neptune CSS is that everyone that want to consume the `Avatar` component sholdn´t have to add whole `components/main.css` but only `neptune-css/dist/avatar.css`.

src/flowNavigation/avatar/Avatar.js
```diff
@@ -1,71 +1,44 @@
...
+  // Reset the errored state when url changes
+  useEffect(() => setImageLoadError(false), [url]);
+
+  return (
+    <div
+      className={classnames('circle circle-responsive circle-inverse', {
+        'tw-avatar--with-image': !isBusinessProfile && renderImage,
+      })}
+    >
+      {!isBusinessProfile ? (
```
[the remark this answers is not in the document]
twrdjpalmer: Using `isPersonalProfile` would change the logic. The original specifically checked that is _wasn't_ a business profile and I don't have enough context to make such a change with confidence.

`const isPersonalProfile = profileType !== BUSINESS` would be misleading too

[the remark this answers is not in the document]
dozortsev: aren't tests give us this confidence? 🙂

anyway I would still write like this:

```javascript
{isBusinessProfile ? (
   <div className="icon icon-bank" />
     ...
) : (
   <div className="tw-avatar__icon-container icon icon-profile">
    ...
)}
```

src/flowNavigation/avatar/Avatar.less
twrdjpalmer: As `Avatar` is an implementation detail of `FlowNavigation` can we delay moving it to `neptune-css` for when the flow nav is moved too?

src/flowNavigation/avatar/Avatar.js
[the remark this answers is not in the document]
twrdjpalmer: I'm not sure what you mean with tests? It's an integration problem more than anything. Take this example:

```jsx
<Avatar url={VALID_URL} profileType="bananas" />
```

Now, we'd render the profile image. If we change the logic to use `isPersonalProfile = profileType === PERSONAL`, we'd render the bank icon instead.

The tenary change makes sense—I'll add that. How would you want to avoid the other parts?

src/flowNavigation/avatar/Avatar.js
twrdjpalmer: Discussed with @andreapiras: going to delay this change as it's non-trivial due to how the flow navigation styles the circle & how the icons are currently misaligned when rendered inside a circle. Will update with the ticket once I've created it

src/flowNavigation/avatar/Avatar.less
twrdjpalmer: Discussed with @andreapiras, going to create a ticket address this in future. Means we can get this improvement in without blocking higher priority work 👍 Will update with the ticket as soon as I've created it

src/flowNavigation/avatar/Avatar.js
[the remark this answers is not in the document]
dozortsev: re. tests

you mention:
> ..The original specifically checked that is wasn't a business profile and I don't have enough context to make such a change with confidence.

tests should give us confidence to do such changes,  so e.g we do refactor of things and then run tests to make sure that component (or e.g business logic) stay the same

----

why would some incorrect value (like `bananas`) might end up passed in the component? do you know cases when it might be possible?

----

thanks that you made a change to ternary operator statement, don't want to make a long discussion from a [minor comment](https://github.com/transferwise/components/pull/245#discussion_r321194430) 🙂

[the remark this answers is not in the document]
twrdjpalmer: Yeah I understand what you mean with the tests, and rather pointing out that changing it to `isPersonalProfile` would inherently change the logic, `isPersonalProfile = profileType !== BUSINESS` is confusing IMO.

The `banana` example is just to demonstrate the point that, it could be any value that isn't `personal` or `business`. I don't know of any cases, no, but that's why I lack the confidence to change the behaviour 😄 

Happy to move forward 👍

PR #248, opened 2019-09-05
Prevent multiple modal close calls

fergusjordantw:
Both the `Dimmer` component and the `Modal` component having event handlers for `onEscape` and `onClick` was causing `onClose` to get called multiple times. 

Not passing through an `onClose` to `Dimmer` from `Modal` fulfils all the different ways in which a modal can be closed and only calls the handler once.


src/modal/Modal.spec.js
```diff
@@ -135,23 +134,34 @@ describe('Modal', () => {
...
+      expect(onClose).toHaveBeenCalledTimes(1);
+    });
+
+    it('only calls close handler once when setting open to false in onClose handler', () => {
+      const onClose = jest.fn(() => {
+        component.setProps({ open: false });
+      });
+
+      component.setProps({ onClose, closeOnClick: true });
+      clickCloseButton();
```
oliverviljamaa: Lovely. Can we expect `onClose` to not have been called before the click?

fergusjordantw: Yep, added it.

PR #252, opened 2019-09-10
rework button, update implementations in docs

sergonius:
- Use `children` instead of a custom prop
- Use the built-in `disabled` and add `loading` prop instead of a generic `state` prop
- Spread any other props people might want to use (`aria-*`, `form*` etc)
- Updated other docs that use `<Button />`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/components",
-  "version": "12.7.1",
+  "version": "12.8.0",
   "description": "TransferWise styleguide components in react",
   "license": "MIT",
   "main": "./build/main.js",
--- a/src/button/Button.js
+++ b/src/button/Button.js
@@ -17,18 +17,9 @@ const Size = {
   Large: 'lg',
 };
 
-const State = {
-  Default: 'default',
-  Loading: 'loading',
-  Disabled: 'disabled',
-};
-
-const Button = ({ label, state, size, type, block, onClick, htmlType }) => {
-  const isLoading = state === State.Loading;
-  const isDisabled = state === State.Disabled || isLoading;
-
+const Button = ({ block, children, disabled, htmlType, loading, size, type, ...rest }) => {
   const classes = classNames(`btn btn-${size}`, {
-    'btn-loading': isLoading,
+    'btn-loading': loading,
     'btn-primary': type === Type.Primary,
     'btn-success': type === Type.Pay,
     'btn-default': type === Type.Secondary,
@@ -37,32 +28,33 @@ const Button = ({ label, state, size, type, block, onClick, htmlType }) => {
   });
 
   return (
-    <button type={htmlType} className={classes} onClick={onClick} disabled={isDisabled}>
-      {label}
-      {isLoading && <span className={classNames('btn-loader', { 'm-l-2': !block })} />}
+    <button type={htmlType} className={classes} disabled={disabled || loading} {...rest}>
+      {children}
+      {loading && <span className={classNames('btn-loader', { 'm-l-2': !block })} />}
     </button>
   );
 };
 
 Button.Type = Type;
 Button.Size = Size;
-Button.State = State;
 
 Button.propTypes = {
   type: Types.oneOf(Object.values(Type)),
   size: Types.oneOf(Object.values(Size)),
-  state: Types.oneOf(Object.values(State)),
+  disabled: Types.bool,
   block: Types.bool,
+  loading: Types.bool,
   onClick: Types.func.isRequired,
-  label: Types.node.isRequired,
+  children: Types.node.isRequired,
   htmlType: Types.oneOf(['submit', 'reset', 'button']),
 };
 
 Button.defaultProps = {
   size: Button.Size.Medium,
   type: Button.Type.Primary,
-  state: Button.State.Default,
+  disabled: false,
   block: false,
+  loading: false,
   htmlType: 'button',
 };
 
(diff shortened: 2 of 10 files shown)


src/button/Button.js
```diff
@@ -17,18 +17,9 @@ const Size = {
...
-  Default: 'default',
-  Loading: 'loading',
-  Disabled: 'disabled',
-};
-
-const Button = ({ label, state, size, type, block, onClick, htmlType }) => {
-  const isLoading = state === State.Loading;
-  const isDisabled = state === State.Disabled || isLoading;
-
+const Button = ({ block, children, disabled, htmlType, loading, size, type, ...rest }) => {
```
mikknarusk: tbh I'm a bit worried about this. Button is probably the most-used component from the library, so making a breaking change like this would mean **a lot** of refactoring. Wouldn't be a problem in most places but I'm afraid that this will force larger repos (mainly frontend-apps) to be stuck with the previous major version as no-one will take on the challenge of refactoring every usage. 

So might be worth researching the usage of this component first to see how many consumers we're forcing choose between a quite large rewrite vs being left behind. Could be a non-issue, but I have a bad feeling considering quite a lot of stuff needed debugging the last time we bumped major versions of `components`  in `frontend-apps`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/components",
-  "version": "12.7.1",
+  "version": "12.8.0",
```
mikknarusk: def a major version bump

src/button/Button.js
```diff
@@ -17,18 +17,9 @@ const Size = {
   Large: 'lg',
 };
 
-const State = {
-  Default: 'default',
```
mikknarusk: what's the reasoning of using explicit props vs a state machine currently? I quite like the state enum approach - don't have to think about weird combinations of props, while it does sacrifice on the flexibility a bit.

sergonius: `Default` serves no real purpose pretty much, `disabled` is already built in/native and we can just add `loading`.
There's currently nothing difficult or any weird prop combinations to take care of here and IMO it makes sense to keep this simple component as close to a regular ol' button.
If we introduce more complex state and can't do with `disabled || loading` etc, we can upgrade.

package.json
```diff
@@ -168,4 +168,4 @@
       "git add"
     ]
   }
-}
\ No newline at end of file
+}
```
oliverviljamaa: 🐱

src/button/Button.js
oliverviljamaa: Is it that used, though?

Also, we could consider a version where we support both, doing `children || label` essentially.

sergonius: I don't have any stats, but I was told it is not really a huge thing yet.
It is also a public facing library and IMO if the changes "make sense" and improve the usability and/or versatility, "we need to refactor in n places" shouldn't stop us from making enhancements.
I find the state machine in particular quite useless for such a trivial component (especially considering that it didn't really simplify anything + actually made the component more verbose).
And no `children` for something that just wants to be a `<button />`?
Also, even if there were a lot of places where refactoring needed to be done, it wouldn't be difficult and would serve a good purpose (if we agree these really are improvements).

package.json
sergonius: 🐹

src/button/Button.js
mikknarusk: I'm not actually arguing against the changes, I think they make sense, especially using `children`. 

I agree that having to refactor stuff shouldn't stop us from improving the components, just considering the fact that we could be essentially sunsetting some consumers. 

Again, could be a non-issue 🤷‍♂

domhorvathtw: Agree with Mikk, not sure how many apps are using it, but would be good to do some research. It's not a really bad breaking change but could be annoying.

src/button/Button.js
domhorvathtw: Both works IMHO. Can't think of any weird state combination either, this approach will actually allow to render a loading button that is disabled. Don't know if that's something we would need, but could be.

src/button/Button.js
```diff
@@ -37,32 +28,33 @@ const Button = ({ label, state, size, type, block, onClick, htmlType }) => {
   });
 
   return (
-    <button type={htmlType} className={classes} onClick={onClick} disabled={isDisabled}>
-      {label}
-      {isLoading && <span className={classNames('btn-loader', { 'm-l-2': !block })} />}
+    <button type={htmlType} className={classes} disabled={disabled || loading} {...rest}>
```
twrdjpalmer: If we're doing a full refactor of the button component like this, I'd like to propose using `type` for the `htmlType` and renaming  `type` to something else

src/button/Button.js
twrdjpalmer: If the change is high risk, could we go through a deprecation cycle? E.g., update the API to what we want it to be, create a wrapper around it which provides the old API and have that throw a warning in development that we'll be changing the api and the consumer should update their code?

Could see it being something like:

```js
import Button from "@transferwise/components"; //old api, throws warning in the console
import { Button } from "@transferwise/components"; // new api, doesn't throw a warning
```

src/button/Button.js
sergonius: I agree that it is not great [, not terrible], but I don't have any terrific ideas either. I only used `htmlType` because Ant did it and it might be "familiar" for others as well.

PR #253, opened 2019-09-10
Standardised month format props in DateInput, DateLookup and FormControl

jacklesliewise:
- Added `MonthFormat` data
- Updated `DateInput` prop types
- Replaced `shortDate` with `monthFormat`


src/common/monthFormat.js
```diff
@@ -0,0 +1,4 @@
+export const MonthFormat = {
```
dozortsev: I would create space for dates, e.g `Dates` 🙂and have constant `MonthFormat` there; eventually it will be a place for other date things

dozortsev: btw, we already have one check `dateUtils` folder

src/dateLookup/DateLookup.docs.js
```diff
@@ -3,6 +3,7 @@ import { DateLookup } from '..';
 import { generateCodeBlock, generateInput, generateState } from '../../docs/utils';
 
 const SIZES = ['sm', 'md', 'lg'];
+const MONTH_FORMATS = ['long', 'short'];
```
dozortsev: let's use the constant ☝️ in docs as well

PR #260, opened 2019-09-13
Add tree-shakeable build for components

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-142.
CR advice: PR is quite big, better to CR commit per commit

## ❓ Context 
Currently components only export umd build which is not treeshakeable causing all the consumer to add extra unnecessary overload 

## 🚀 Changes
This PR introduces:
1) Es6 exported version of components
2) Separated css for component that will be auto-importedautomatically when component gets imported

## 💬 Considerations 
Ideally all code reviewers should test this library in an existing project and verify that bundle works as expected and that the bundle size gets reduced. Step for testing:
1) `cd components/`
2) `git checkout DS-142-treeshake`
3) `npm install` and `npm run build` (now you should have a lib folder please have a look and feel free raise concerns)
4) `npm link`
5) `cd your-app`
6) build your app (please measure the size (you could use `ls -lh` command) and DM to me)
7) `npm link "@transferwise/components"`
8) build your app again (please measure the size and DM to me)
9) check bundle(s) for unused components, they shouldn't be there
10) start your app and verify it works as expected

**Tree Shaking results (bundle(s) decrease) in some apps**: https://docs.google.com/spreadsheets/d/1vZKDZO377O43ocrq89GZc4Gg00rcgTXrZmKvxxoATeo/edit#gid=0

## ✅ Checklist


rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// Babel
+const babelrc = false;
+const babelPlugins = [
+  '@babel/plugin-proposal-class-properties'
+];
+const exclude = 'node_modules/**';
+const presets = [
+  ["@babel/env"],
```
dozortsev: is it actually necessary to wrap this in an array?

package.json
```diff
@@ -37,6 +38,7 @@
     "bootstrap": "github:transferwise/bootstrap#v5.18.15",
     "camelcase": "^5.2.0",
     "case-sensitive-paths-webpack-plugin": "2.2.0",
+    "cpx": "^1.5.0",
```
dozortsev: why do we need this?

package.json
```diff
@@ -134,11 +135,17 @@
...
   "scripts": {
-    "start": "node scripts/start.js",
-    "build": "node ./emotionFix.js && node scripts/build.js",
+    "dev": "npm-run-all dev:compile-less --parallel dev:webpack dev:watch-less  ",
+    "dev:compile-less": "gulp lessCompiler",
+    "dev:webpack": "node scripts/start.js",
+    "dev:watch-less": "gulp lessWatcher",
+    "build": "npm-run-all build:*",
+    "build:clean": "rm -rf lib build",
+    "build:emotionFix": "node ./emotionFix.js",
```
dozortsev: please keep consistent naming for commands, `emotion-fix`

package.json
```diff
@@ -81,6 +81,7 @@
     "lint-staged": "^9.0.2",
     "lodash.times": "^4.3.2",
     "mini-css-extract-plugin": "0.5.0",
+    "npm-run-all": "^4.1.5",
```
dozortsev: why not just use simple `npm`? 🙄

rollup.config.js
```diff
@@ -0,0 +1,66 @@
...
+
+// Rollup
+const input = 'src/index.js';
+const external = ['react', 'react-dom', 'prop-types'];
+const namedExports = {
+  'node_modules/@transferwise/formatting/dist/formatting.js': [
+    'formatAmount',
+    'formatMoney',
+    'formatDate',
+    'formatNumber',
```
taaviaasver: Not sure why is this list here, but - is that scalable? What if we add `formatWhatever` to `formatting`?

package.json
```diff
@@ -111,17 +136,26 @@
     "react-transition-group": "^4.2.1"
   },
   "scripts": {
-    "start": "node scripts/start.js",
-    "build": "node ./emotionFix.js && node scripts/build.js",
```
taaviaasver: Maybe some dependencies could be removed from the project? (since `build.js` was deleted)

package.json
andreapiras: Is just a matter of syntax really. I found this more readable and concise ```npm-run-all build:*```

rollup.config.js
andreapiras: This is a problem with Rollup and default exports. 
image.
To be honest I would not care too much as the umd is for legacy code and should be removed in favour of the tree-shakeable version

package.json
andreapiras: Yes, onve we'll move components to monorepo and switch to the new docs we can get rid of webpack completely. For now is still needed to spin up the old dev env and to build the old docs

package.json
```diff
@@ -24,6 +27,7 @@
     "@svgr/webpack": "4.1.0",
     "@typescript-eslint/eslint-plugin": "1.6.0",
     "@typescript-eslint/parser": "1.6.0",
+    "autoprefixer": "^9.6.1",
```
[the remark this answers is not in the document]
andreapiras: this gets added automatically when install by alphabetical sorting :)

rollup.config.js
andreapiras: @taaviaasver I added a comment to the code and a reference to some docs.
As I said I would not worry to much about this as it can be already considered legacy code. The bundle produced with rollup is umd in this case and is not tree-shakeable. The lib one is the hot one that gets automatically imported when u point to @transferwise/components because of the module field in package.json

.babelrc
```diff
@@ -7,10 +7,16 @@
         "corejs": "3.0.0",
         "targets": {
           "browsers": ["last 2 versions", "safari >= 9", "ie >= 11"]
-        }
+        },
+        "modules": false
       }
     ]
   ],
+  "env": {
```
andreapiras: Jest still need files to be transpiled as cjs in order to work

.babelrc
```diff
@@ -7,10 +7,16 @@
         "corejs": "3.0.0",
         "targets": {
           "browsers": ["last 2 versions", "safari >= 9", "ie >= 11"]
-        }
+        },
+        "modules": false
```
andreapiras: This prevents babel to transpile es6 to cjs . Cjs is not treeshakeable

gulpfile.js
```diff
@@ -0,0 +1,28 @@
+var gulp = require('gulp');
```
andreapiras: This is a temporary less compiler. This will be removed once we switch fully to postcss syntax

package.json
```diff
@@ -1,20 +1,25 @@
 {
   "name": "@transferwise/components",
-  "version": "13.0.4",
+  "version": "14.0.0",
   "description": "TransferWise styleguide components in react",
   "license": "MIT",
   "main": "./build/main.js",
+  "module": "./lib/index.js",
+  "style": "./build/main.css",
+  "sideEffects": false,
```
andreapiras:
```In a 100% ESM module world, identifying side effects is straightforward. However, we aren't there just yet, so in the mean time it's necessary to provide hints to webpack's compiler on the "pureness" of your code.

The way this is accomplished is the "sideEffects" package.json property.```

https://webpack.js.org/guides/tree-shaking/

File: src/accordion/Accordion.js [hunk 12]
```diff
@@ -3,7 +3,7 @@ import Types from 'prop-types';
 
 import AccordionItem from './AccordionItem';
 
-import './Accordion.less';
+import './Accordion.css';
```
andreapiras: With this change basically when we'll import 
``` import {Accordion} from '@transferwise/components```
the component will import Accordion.css so we won't have to include the whole main.css anymore

File: src/common/dateUtils/moveToWithinRange/moveToWithinRange.js [hunk 13]
```diff
@@ -1,4 +1,4 @@
-import { isWithinRange } from '..';
+import { isWithinRange } from '../isWithinRange';
```
andreapiras: Removes circular dependencies

PR #261, opened 2019-09-16
Tabs

sergonius:
## 💬 Feel free to propose ideas for props etc

Currently, we're thinking of passing an `tabs` prop through to the component which contains the content for the "tab panel" and the title for the tab itself. i.e.

```javascript
<Tabs 
    tabs=[
        {
            title: 'a title',
            content: <PanelContent />
        },
        {
            title: 'a second title',
            content: <DifferentPanelContent />
        }
    ]
/>
```

## ✅ Checklist


src/tabs/Tab.js
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
```
oliverviljamaa: Why not `classNames` like everywhere else?

fergusjordantw: The first bit of stuff @sergonius pulled in from a standard tabs library, so some stuff just hasn't had surgery performed on it yet

gusso:
Did you consider going the route of seeing the change live of the dragging effect, and have the tab content snap to its appropriate place onTouchEnd instead of the gesture?

fergusjordantw:
> Did you consider going the route of seeing the change live of the dragging effect, and have the tab content snap to its appropriate place onTouchEnd (like we do on the mobile app) instead of the gesture?

We will have that (and do as of the latest commit - although its not yet tested on an actual phone, just the chrome simulator). Have just been slowly tidying up this stuff one small step at a time.

gusso:
NEAT! Ok, some feedback:

- We should lock horizontal scrolling when the user starts a vertical scroll. If the user starts a vertical scroll within the content of the tab, even minor horizontal deltas trigger the horizontal scroll as well, which makes it look a bit janky.
- The horizontal scroll feels a bit unnatural, I wonder if it wouldn't be possible to use the native horizontal scroll of the browser and just use JS to snap to the correct tab when letting go?

src/tabs/Tabs.less
```diff
@@ -0,0 +1,82 @@
+@import '~bootstrap/less/variables.less';
```
dozortsev: could you please create this in [neptune/css](https://github.com/transferwise/neptune/tree/master/packages/css) (monorepo)?

fergusjordantw: The entire tabs CSS? Definitely can do, just wanted to keep it here at the moment for a faster development cycle.

fergusjordantw:
@gusso 
> We should lock horizontal scrolling when the user starts a vertical scroll. If the user starts a vertical scroll within the content of the tab, even minor horizontal deltas trigger the horizontal scroll as well, which makes it look a bit janky.

Definitely. We focused so far on disabling vertical scroll when swiping horizontally, but will look at doing the opposite of that as well for a smoother experience.

> The horizontal scroll feels a bit unnatural, I wonder if it wouldn't be possible to use the native horizontal scroll of the browser and just use JS to snap to the correct tab when letting go?

How have you been testing it? We've been developing it against the same currency calculator swiping experience in the apps to try and be in line with what the app guys are doing. We looked at using normal horizontal scrolling (as well as using https://css-tricks.com/practical-css-scroll-snapping/) but the experiences were both horrible.

src/tabs/Tabs.less
dozortsev: nice,

soon we move to [monorepo](https://github.com/transferwise/neptune) and devs will have fast development cycle. 🙂

src/tabs/Tabs.js
```diff
@@ -0,0 +1,411 @@
...
+    const restrictedVelocity =
+      (Number.isNaN(remainingContainerToTravel) ? 0 : remainingContainerToTravel) *
+      Math.min(10 * Math.E, lastSwipeVelocity * 10 * Math.E);
+
+    return (
+      <div
+        onTouchStart={changeTabOnSwipe && this.handleTouchStart}
+        onTouchEnd={changeTabOnSwipe && this.handleTouchEnd}
+        onTouchMove={changeTabOnSwipe && this.handleTouchMove}
+        className="tabs"
```
mikknarusk: is `tabs` maybe too generic? I'm afraid of this component's styles affecting other stuff in the same page or vice versa

mikknarusk: Also I've found it useful to have the component pass down a `className` prop to the root DOM element. So you'd be able use it as such without having to wrap it in a div:
```
<Tabs className="m-t-2" {...props} />
```

src/tabs/Tab.js
```diff
@@ -0,0 +1,53 @@
...
+  selected: false,
+};
+
+Tab.propTypes = {
+  children: PropTypes.node.isRequired,
+  disabled: PropTypes.bool,
+  selected: PropTypes.bool,
+  id: PropTypes.string.isRequired,
+  panelId: PropTypes.string.isRequired,
+  handleKeyDown: PropTypes.func.isRequired,
```
mikknarusk: what do you think of `onKeyDown` being the prop name? Just to align with usual event listener lingo

src/tabs/TabList.js
```diff
@@ -0,0 +1,18 @@
...
+      {children}
+    </ul>
+  );
+};
+
+TabList.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.object, PropTypes.array]).isRequired,
+};
+
+TabList.tabsRole = 'TabList';
```
mikknarusk: what does this do?

fergusjordantw: Holdover from https://github.com/reactjs/react-tabs which is where sergo scaffolded the original structure from. Will remove.

src/tabs/Tabs.less
```diff
@@ -0,0 +1,78 @@
...
+@import '~@transferwise/neptune-css/src/less/variables/_colors.less';
+@import '~@transferwise/neptune-css/src/less/variables/_typography.less';
+@import '~@transferwise/neptune-css/src/less/variables/_transitions.less';
+
+@transition-timing-function: ease-in-out;
+@transition-duration: 0.3s;
+@short-transition-duration: 0.15s;
+@line-thickness: 3px;
+
+.tabs {
```
mikknarusk: @fergusjordantw was this moved to neptune?

src/tabs/Tabs.js
```diff
@@ -0,0 +1,411 @@
...
+  swipedRightToLeft,
+  swipeShouldChangeTab,
+  getVelocity,
+} from './utils';
+
+import './Tabs.less';
+
+const MIN_INDEX = 0;
+
+const tabFilter = tab => !tab.disabled;
```
mikknarusk: can we call it something more relevant?

src/tabs/Tabs.less
fergusjordantw: Not yet, have been waiting to see where everything ends up with the monorepo to add it closer to when it's ready to merge.

src/tabs/Tabs.js
fergusjordantw: `enabledTabsFilter` do it for ya?

mikknarusk: yes

src/tabs/Tabs.js
```diff
@@ -0,0 +1,411 @@
...
+      Math.max(selected - 1, MIN_INDEX),
+      Math.min(selected + 1, this.MAX_INDEX),
+    );
+
+    return nextSelected;
+  };
+
+  swipedOverHalfOfContainer = difference => difference / this.containerWidth >= 0.5;
+
+  calculateApplicapleDragDifference = (currentSelected, start, end) => {
```
mikknarusk: `Applicable`?

fergusjordantw: Hahah, yes.

src/tabs/Tabs.js
```diff
@@ -0,0 +1,411 @@
+import React from 'react';
```
mikknarusk: would love if we could extract even more of this file into either `utils` or some other file in order to separate all the swiping logic from just rendering stuff as much as possible

fergusjordantw:
> hey @fergusjordantw and @sergonius
> 
> found a bug for mobile web clients

@dozortsev I think this is just when you resize the screen (we don't have a resize listener) and if you refresh it'll be fine. This said - do you think we should add a resize handler?

dozortsev:
@fergusjordantw ah indeed,

no, no need resize handler

src/tabs/Tab.js
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+const Tab = ({ children, id, disabled, panelId, selected, handleKeyDown, ...attributes }) => {
```
dozortsev: Hey Sergo,

I a bit disagree exposing interface of components for attributes in this way (`...attributes`)

- this makes the interface of component not explicit, which will **drive inconsistency** of the way people will use it, e.g we might won’t people use `form` attr in `<Button>`, just because we will have our own `<Form>` and they will be linked in a different way to satisfy our needs and if we some something like that we want have consistency in they way people use them (maybe bad example but I hope you understand me)

- re. **aria**, this won’t help drive it, because people will forget add aria-* attrs plus sometime they just don’t know what to add, soon DS team will add/expose aria-* attrs to all of our components to ensure accessibility

- **tests** & **documentation**, spreading props (`...attributes`) makes a hole as in tests as in documentation, we want props & state of our components be well known and well covered

we would go with strict and explicit way of defining the interface of our components, this we will allow us drive consistency, a11y and I believe some other stuff. If consumer missing some prop, they should (discuss #design-system and) contribute 🙂

If you still disagree we open for discuss and if this really makes sense then we add this to other components as well (again because of consistency) but not now, it’s not a priority for us atm, sorry 🙂

So please remove this from `<Tab>`, `<Tabs>` components and from [`<Button>` one](https://github.com/transferwise/components/blob/master/src/button/Button.js#L20).

package.json
```diff
@@ -133,6 +133,7 @@
     "lodash.clamp": "^4.0.3",
     "lodash.debounce": "^4.0.8",
     "react-app-polyfill": "^1.0.1",
+    "react-spring": "^8.0.27",
```
andreapiras: Has this been discussed in the guild? We currently use react-transition-group. I don't have any preference between one or  the other. I'd keep contained the introduction of new packages. so if we decide to go with spring I have nothing against but let's make sure we discuss that in the guild

andreapiras: Also looks like doesn't support ie11? or Am I wrong
image

fergusjordantw: We use the `commonJS` export so it should be fine. As for using this over `react-transition-group`, I don't really think it's a case of one or the other. AFAIK `react-transition-group` doesn't do the same gradual easing that `react-spring` does. `react-spring` easily looked like the best option to handle this in a manageable way when @sergonius and I looked for solutions for the easing. And it's a small enough package that it shouldn't be a major issue including it (especially with the upcoming tree-shaking work).

src/tabs/Tab.js
fergusjordantw: Sounds good

src/tabs/Tabs.js
mikknarusk: agreed on that.
@dozortsev do you have an opinion on this maybe?

package.json
dozortsev: can we use react-transition-group lib (which is already added) instead of this lib?

src/tabs/Tab.js
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+const Tab = ({ children, id, disabled, panelId, selected, onKeyDown, ...attributes }) => {
```
dozortsev: we agreed to remove prop spreading for _Tab_

fergusjordantw: Yep - just haven't got to it yet.

image

Github doing this shit is making it difficult to keep track of all the different comments/changed that need to be made.

package.json
dozortsev: it would be great to have as less as possible of such libraries in components lib and try to re-use features of ones which are already added and used in other components, because of consistency - diff libs brings diff approaches for animations and stuff like that...

fergusjordantw: @dozortsev 👍 I'm going to be discussing it tomorrow with @andreapiras

fergusjordantw: Talked with Andrea, we're all good to move forward barring any major push back here https://transferwise.slack.com/archives/C2SGDKP4Y/p1571999036208900

PR #299, opened 2019-10-05
Correctly render a disabled field when prop is based

alifarah9:
Correctly render a disabled field when prop is based.

#First PR on github


src/field/Field.js
```diff
@@ -279,7 +279,7 @@ export default class Field extends Component {
   }
 
   render() {
-    const { name, field, value, locale, label } = this.props;
+    const { name, field, value, locale, label, disabled } = this.props;
```
dozortsev: we pass this prop via `field` prop https://github.com/transferwise/components/blob/master/src/field/Field.js#L60

dozortsev:
@alifarah9 thanks for interest to our components library 

but this fix is redundant 🙂

PR #300, opened 2019-10-09
fix searchValue prop propagation in field and formControl

gittymcgit:
## ❓ Context
Form Control component wouldn't propagate searchValue to Select. Making it impossible to create searchable selects using formControl or Field componets

## ✅ Checklist

- [ ] All changes are covered by tests


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v13.0.5
+## Add searchValue prop ro formControl
```
dozortsev: tiny spelling mistake :)

StevePole:
What's the use case?  Never came across a need for this in the Angular library.

gittymcgit:
@StevePole If you can't provide searchValue to select, search doesn't work at all. We need search for country selector

PR #318, opened 2019-10-11
Add TransferWise ESLint config, update related dependencies

oliverviljamaa:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/BORAT-3202

## ❓ Context  
This package was not using the [@transferwise/eslint](https://github.com/transferwise/eslint-config) package that since version 4 does not require `babel-eslint` parser or any config (Prettier, Airbnb), allowing us to unify the code style.

## 🚀 Changes 
* TransferWise ESLint config is used
* rules and parser already applied there are removed
* packages already used through there are removed
* Prettier config from that package is used (to support IDEs without ESLint auto-fixing)
* `prettier-eslint` is removed, as Prettier rules are applied through ESLint
* pre-commit hooks are fixed
* new rules that came with new versions of configs are ignored (when questionable) or fixed (when certainly correct)

## 💬 Considerations 
Best read commit by commit, as there are 2 auto-fix commits.

## ✅ Checklist


src/accordion/Accordion.spec.js
```diff
@@ -28,7 +28,7 @@ describe('Accordion', () => {
           },
           {
             title: 'c',
-            content: <h1>I'm a h1 element</h1>,
+            content: <h1>I&apos;m a h1 element</h1>,
```
OFFLlNE: interesting

src/accordion/AccordionItem/AccordionItem.js
```diff
@@ -21,7 +21,7 @@ class AccordionItem extends PureComponent {
     const { title, content } = this.props;
     const { isOpen } = this.state;
     return (
-      <Fragment>
+      <>
```
OFFLlNE: Not sure if we should do that. At least in Homepage project they caused some issues.

src/dateInput/DateInput.view.spec.js
```diff
@@ -236,27 +234,27 @@ describe('Date Input Component', () => {
     });
 
     it('corrects days too high in February', () => {
-      const component = shallow(<DateInput {...props} />);
+      const comp = shallow(<DateInput {...props} />);
```
OFFLlNE: new conf is with `comp`? I liked component more :D

src/dateLookup/DateLookup.keyboardEvents.spec.js
```diff
@@ -149,7 +149,7 @@ describe('DateLookup (keyboard events)', () => {
     component.simulate('keyDown', fakeKeyDownEventForKey(key));
   };
 
-  const onChangeCalledWith = date => {
-    expect(props.onChange).toHaveBeenLastCalledWith(date);
+  const onChangeCalledWith = d => {
```
OFFLlNE: would prefer `date` here

src/formControl/FormControl.spec.js
```diff
@@ -353,37 +352,37 @@ describe('FormControl', () => {
     });
   });
 
-  function testAutoComplete(controlType, controlSelector) {
-    let control;
+  // function testAutoComplete(controlType, controlSelector) {
```
OFFLlNE: why?

oliverviljamaa: The tests using it are commented out, so I didn't want to just delete this one either.

src/dateLookup/DateLookup.keyboardEvents.spec.js
oliverviljamaa: Same here, but `date` is already defined in the scope.

src/dateInput/DateInput.view.spec.js
oliverviljamaa: Me too, but `component` is already defined in the scope.

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: Interesting, can you explain what the issues were? The shorthand is semantic, requires one less import, and I haven't heard of any behaviour differences.

src/accordion/Accordion.spec.js
oliverviljamaa: Yeah 🤷‍♂

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: The only issue I can see is if someone's using an older React version (Fragment shorthand is supported from 16.2, I believe).

OFFLlNE: @KarlEST Probably knows best. I just remember this was causing big problems and this was the cause. Might just be connected to apps served from FES, but wanted to comment just in case

src/formControl/FormControl.spec.js
OFFLlNE: fair

src/dateInput/DateInput.view.spec.js
OFFLlNE: 😭

src/dateLookup/DateLookup.keyboardEvents.spec.js
OFFLlNE: 😭

src/accordion/AccordionItem/AccordionItem.js
oliverviljamaa: Yup, it's good feedback!

src/dateInput/DateInput.view.spec.js
dozortsev: but you don't to define it, just re-assign:

```javascript
component = shallow(<DateInput {...props} />);
```

dozortsev: just like we do in [previous tests](https://github.com/transferwise/components/pull/318/files#diff-c209a7ef07c91ad2dc1461707c73bab2R223)

src/dateLookup/DateLookup.events.spec.js
```diff
@@ -113,9 +113,9 @@ describe('DateLookup (events)', () => {
 
   it('updates selected date and closes', () => {
     component.setState({ open: true });
-    const date = new Date(2018, 11, 28);
-    component.instance().handleSelectedDateUpdate(date);
-    expect(props.onChange).toHaveBeenCalledWith(date);
+    const d = new Date(2018, 11, 28);
```
dozortsev: same here

src/dateLookup/DateLookup.keyboardEvents.spec.js
dozortsev: 😭

src/definitionList/DefinitionListItem.js
```diff
@@ -62,6 +62,10 @@ DefinitionListItem.propTypes = {
     title: Types.string.isRequired,
     type: Types.string.isRequired,
     width: Types.string,
+    tagClassName: {
+      h3: Types.bool,
+    },
+    values: Types.arrayOf(Types.shape({ value: Types.any })),
```
dozortsev: why this necessary? 

why `h3` prop is boolean?

src/formControl/FormControl.spec.js
dozortsev: and why those tests commented out?

src/typeahead/typeaheadInput/TypeaheadInput.js
```diff
@@ -135,6 +135,8 @@ export default class TypeaheadInput extends Component {
           </div>
         </div>
       );
-    } else return this.renderInput();
+    } else {
+      return this.renderInput();
+    }
```
dozortsev: minor thing: here we don't need `else` clause, so could be slightly simplified:
```javascript
}
return this.renderInput();
```

src/formControl/FormControl.spec.js
oliverviljamaa: I do not know.

src/dateInput/DateInput.view.spec.js
oliverviljamaa: I don't think that's better and didn't want to change behaviour 🤔

What do you think?

src/definitionList/DefinitionListItem.js
oliverviljamaa: The ESLint config required prop types for these values and I added these types based on usage (yup, `h3` is a `boolean` prop).

src/dateInput/DateInput.view.spec.js
dozortsev: hard to know w/o context, but perhaps worth trying, but anyway...

src/formControl/FormControl.spec.js
dozortsev: but it shows like it's your changes in this PR, how come you don't know? 🙂

src/typeahead/typeaheadInput/TypeaheadInput.js
oliverviljamaa: Thanks, done in b115030b026a00a2daf5e996d61ed19987779f02.

PR #320, opened 2019-10-16
Improved phone number search

andreapiras:
## 📎 Jira ticket
https://transferwise.atlassian.net/browse/DS-145

## ❓ Context  
The current phone number atm only allows to search prefixes based on Iso2 and Iso3. With this PR the search will be extended to prefixes numeric values

## 🚀 Changes 
This Pr extends the search to be performed also on numeric prefix values

## ✅ Checklist


src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -70,9 +71,14 @@ class PhoneNumberInput extends PureComponent {
...
   getSelectOptions = () => {
-    const filteredOptions = this.state.searchQuery
-      ? filterOptionsForQuery(countries, this.state.searchQuery)
-      : countries;
+    const { searchQuery } = this.state;
+    const filteredOptions = searchQuery ? filterOptionsForQuery(countries, searchQuery) : countries;
+    // If searchQuery is a number sort the options by phone values
+    if (searchQuery && !Number.isNaN(convertStringToNumber(searchQuery))) {
+      filteredOptions.sort(
+        (a, b) => convertStringToNumber(a.phone) - convertStringToNumber(b.phone),
```
skywritergr: This is confusing the bejezus out of me. Can we add some comment or something? Why are we doing it?

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -70,9 +71,14 @@ class PhoneNumberInput extends PureComponent {
   };
 
   getSelectOptions = () => {
-    const filteredOptions = this.state.searchQuery
-      ? filterOptionsForQuery(countries, this.state.searchQuery)
-      : countries;
+    const { searchQuery } = this.state;
+    const filteredOptions = searchQuery ? filterOptionsForQuery(countries, searchQuery) : countries;
```
dozortsev: why we need do this casting and sorting? 🙂

did you try add `phone` prop here? https://github.com/transferwise/components/blob/59d39f1e7bf3083648ad3d2464e11388a528ab61/src/phoneNumberInput/utils/isOptionAndFitsQuery/index.js#L10

```javascript
contains(option.phone, query) ||
```

andreapiras: Hey Anton thanks for ur CR. We need sorting because as part of the new requirements is you search with a string we want results to be ordered by ISO3 alphabetical,if we search by 'number' we want numbers to be ascending

andreapiras: The casting is needed to convert numerical strings into numbers so we can recognize if someone is searching by "USA" or "1"

src/phoneNumberInput/PhoneNumberInput.spec.js
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
```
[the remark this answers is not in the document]
andreapiras: because I need to test that when typing +1 on the select the results are coming sorted ascending. I could have tested using the internal state of the component but this is not a good practice as what I really need is to provide an input and check the output without checking the internal implementation (that can always change)

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -70,14 +77,26 @@ class PhoneNumberInput extends PureComponent {
...
-    const filteredOptions = this.state.searchQuery
-      ? filterOptionsForQuery(countries, this.state.searchQuery)
-      : countries;
+    const { searchQuery } = this.state;
+    let filteredOptions = this.listSortedByISO3;
+    // If searchQuery is a number sort the options by phone values
+    if (isStringNumeric(searchQuery)) {
+      filteredOptions = filterOptionsForQuery(this.listSortedByPhone, searchQuery);
+    } else {
+      filteredOptions = filterOptionsForQuery(this.listSortedByISO3, searchQuery);
```
dozortsev: why here we need assign `this.listSortedByISO3` if ultimately we re-assign it, could this simplified using `? :`?

```javascript
const filteredOptions = filterOptionsForQuery(
  isStringNumeric(searchQuery) ? this.listSortedByPhone : this.listSortedByISO3,
  searchQuery); 
```

src/phoneNumberInput/PhoneNumberInput.js
```diff
@@ -70,14 +77,26 @@ class PhoneNumberInput extends PureComponent {
...
+    }
 
     return filteredOptions.map(option => {
       const { phone, iso3, iso2 } = option;
+      let note = '';
+
+      if (iso3) {
+        note = isArray(iso3) ? iso3.join(', ') : iso3;
+      } else if (iso2) {
+        note = isArray(iso3) ? iso3.join(', ') : iso3;
```
dozortsev: there is no diff in 94 and 96 lines, if this intentional then this can be simplified:

```
if (iso3 || iso2) {
..
```

src/phoneNumberInput/utils/groupCountriesByPrefix/groupCountriesByPrefix.js
```diff
@@ -0,0 +1,26 @@
+import { isArray } from '../../../common/validation/type-validators';
```
dozortsev: please rename this file to `index.js`

dozortsev: or create `index.js` in same folder for exposing it the function

src/phoneNumberInput/PhoneNumberInput.js
andreapiras: Line 96 is wrong good spot

PR #325, opened 2019-10-23
WIP Introduce JsonSchemaForm

StevePole:
# Conflicts:
#	docs/Docs.js
#	src/index.js
#	src/ssr.spec.js


src/jsonSchemaForm/validation/index.js
```diff
@@ -0,0 +1,19 @@
+import angular from 'angular';
```
Tankenstein: Guessing this shouldn't be here

PR #327, opened 2019-10-28
Accept node as text for Snackbar

hendrikmolder:
## ❓ Context  
We use `react-intl` in our application and would like to pass translated strings via `<FormattedMessage />` component of `react-intl`.

## 🚀 Changes 
- change `Snackbar`s proptypes to accept `node` as well as `string`

## ✅ Checklist


src/snackbar/Snackbar.js
```diff
@@ -101,7 +101,7 @@ Snackbar.propTypes = {
     label: PropTypes.string.isRequired,
     onClick: PropTypes.func.isRequired,
   }),
-  text: PropTypes.string.isRequired,
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
oliverviljamaa: `node` allows `string`, so we can just do:
```suggestion
  text: PropTypes.node.isRequired,
```
😍

src/snackbar/Snackbar.spec.js
```diff
@@ -46,6 +46,26 @@ describe('Snackbar', () => {
     expect(createPortal).toBeCalledWith(<Snackbar {...props} />, document.body);
   });
 
+  it('accepts node as text', async () => {
```
oliverviljamaa: Following up on the previous comment, as node can _be_ a string, we could clarify:
```suggestion
  it('accepts element as text', async () => {
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/components",
-  "version": "15.1.1",
+  "version": "15.1.2",
```
oliverviljamaa: The `test` command is failing, as we're missing a `CHANGELOG` item.

oliverviljamaa:
Can we make the docs use an element instead of a string as well, so we'd cover the slightly more complex scenario?

